PR #56, opened 2020-03-12
43 data series color

whatisgalen:


whatisgalen:
"missing semicolon" fix appears to be a bug with stickler, after several attempts was unable to fix but no clear impact on code running. Also note that while "colorPicker is defined by never used", in its context of the viewmodel, it is being used in the corresponding template `afs-reader.htm` thus the error can be ignored.

Issue #44, opened 2020-03-02
Add the selected file name to the top of the chart renderers.


chiatt:
The selected file name should be visible from the top of the renderer

dwuthrich:
meets requirement

PR #65, opened 2020-03-13
uses ko obs to set filename from tile when available, key off file-li…

whatisgalen:
…st datatype node in graph, re #44


whatisgalen:
lint error can be ignored, syntax used from:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

PR #121, opened 2020-05-06
fix functions, anno cards for PhysThing

azaroth42:
Reorder and relabel the cards for annotation / identifier assignment in the Physical Thing model.
Fix the description function reference.


azaroth42:
Prefer someone else to hit merge please :)

PR #131, opened 2020-05-08
122 iiif fn

whatisgalen:
creates a new function that copies files (of specified acceptable types) to a local cantaloupe directory for the purpose of locally hosting iiif server


whatisgalen:
stickler error inaccurate, merging

Issue #221, opened 2020-07-13
New Collection workflow is broken at last step


dwuthrich:
Arches throws an error when pressing the "Save and continur" button in the New Collection workflow:

![Screen Shot 2020-07-13 at 1.57.12 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/c0021e7f-780a-4170-a2eb-98b7086d612c)

PR #235, opened 2020-07-16
Fixes to support related resources changes

chiatt:
Additional fixes made my @apeters to support changes to the resource instance datatype. re #221


chiatt:
Changes work as expected

PR #250, opened 2020-07-18
Extends rather than overwrites arches RENDERERS setting

chiatt:
Extends rather than overwrites arches RENDERERS setting. re archesproject/arches#6240


njkim:
It works as expected

Issue #267, opened 2020-08-27
Create sample alternative workflow buttons


whatisgalen:
Per https://github.com/archesproject/arches/issues/6248

A developer can now use/insert alternative buttons for a specified workflow step. A template is needed.

PR #268, opened 2020-08-27
267 altbuttons workflow

whatisgalen:
re #267 

Provides template for alternative workflow buttons per PR done for https://github.com/archesproject/arches/issues/6248


chiatt:
This was a good first approach, but we are now moving workflow and card buttons into the workflow footer. re archesproject/arches#7040

PR #270, opened 2020-08-28
Pagination now works in the collection workflow

chrabyrd:
This adds pagination functionality. 

UX seen here: https://recordit.co/JXOm60nJfX


chrabyrd:
Temporarily closing while refactoring search components

PR #271, opened 2020-09-01
144 cbyrd add pagination behavior to collection workflow

chrabyrd:
This gives the collection workflow a functioning search, using the generic components from arches. 

NOTE: This depends on https://github.com/archesproject/arches/pull/6603 !


apeters (approved):
paging now works as expected

PR #272, opened 2020-09-05
263 merge collection research workflows

### LINKED ISSUE #263: Merge the Research Activity and New Collection worflows
Currently AfS has a separate workflow for creating a new research activity and a new collection.  These workflows should be merged, with the collection workflow slotted into the "select set" step of the new research activity.

Note:
The collection/set workflow should hide the "name collection" step, and automatically name the new collection as follows:
Study set for <name of activity> (added by user in the first step of the workflow)

whatisgalen:
This PR needs changes (or something comparable that would achieve the same functtionality) from https://github.com/archesproject/arches/pull/6614

Essentially, this fixes #263 by merging the collection workflow into a single step slotted into the New Research Activity workflow


chiatt (changes requested):
Hey @whatisgalen, unfortunately this breaks in the latest because of the pagination added to the object search step. Changes to the object search step will need to be incorporated into this one. The changes are made in several PRs:

https://github.com/archesproject/arches-for-science-prj/pull/270/files
https://github.com/archesproject/arches-for-science-prj/pull/271/files
https://github.com/archesproject/arches-for-science-prj/pull/275/files

This diff might also help: https://github.com/archesproject/arches-for-science-prj/compare/215_file_iiif_instruction...master

chiatt (changes requested):
@whatisgalen This works much better than before, but it looks like nothing is added to the state, so when a user reloads the collection step, the tile data is not reloaded.

chiatt (changes requested):
Also - it looks like the tile value is sometimes a KO mapping object. That probably happened before you started work on this, but it would be good to address it as part of this PR.  The following lines should be updated to use koMapping like so: koMapping.toJSON(self.value()), so that data is saved to the tile properly.

https://github.com/archesproject/arches-for-science-prj/blob/8fd9f275133857c82deadb9ed08efc7386003bba/afs/media/js/views/components/workflows/research-collection-step.js#L198

https://github.com/archesproject/arches-for-science-prj/blob/cbfd23c3062a7824b56f079b93f5e4cb4d30e5cd/afs/media/js/views/components/workflows/object-search-step.js#L157

chiatt:
@whatisgalen I saw that you  committed some of the work from archesproject/arches#6614 into core arches. I was hoping this PR might work with those changes, but not yet. You may have already known that, though.

chiatt (approved):

PR #283, opened 2020-11-11
Support multiple images in a manifest, re: Arches #6838

njkim:
Add the support for multiple images in a manifest, re #6838
Support multiple canvases to add more than one image to a manifest.
Also, change the iiif_manifest model to add the manifest json column


chiatt (approved):
Works as expected

PR #297, opened 2020-12-24
284 create sample workflow

chrabyrd:
This adds the basic plugin data, as well as steps built around new localStorage logic


chiatt (approved):
Works as expected

Issue #299, opened 2021-01-13
Set workflow template as tabbed for all workflows


chiatt:
@chiatt commented on [Tue Jan 12 2021](https://github.com/archesproject/arches/issues/7001)

All workflows in AfS should use the 'tabbed' template.

PR #300, opened 2021-01-13
all workflows changed to tabbed workflows #299

chrabyrd:
@chiatt should we consider removing `icon` values from steps as part of this ticket?


chiatt (approved):
Works as expected

PR #308, opened 2021-01-20
New project workflow , #281

njkim:
The "New Project" workflow was added, replacing the current "Research Activity" workflow. #281


chiatt (approved):
Works as expected

Issue #310, opened 2021-01-26
Some Digital Resources do not load because of an error in the file to iiif function.


chiatt:
When loading csv business data, only some of the Digital Resource instances successfully load. Those resources that have `file-list` values fail due to the following error in the file-to-iiif function:
```
Mon Jan 25 16:33:28 2021 WARNING: 2021-01-25 18:33:27.943182: WARNING: failed to index document in resource: 945dfb90-59eb-454a-bbba-c5c4f03c50dd on line 12. Exception detail:
local variable 'file_url' referenced before assignment
```

chiatt:
Resolved in #311

PR #311, opened 2021-01-26
Exit the iiif function for unsupported files, #310

njkim:
Stop and exit the function when the file formats are unsupported. #310 
To avoid the function failing when it expects the image in the iiif server.


chiatt (commented):
Works as expected

PR #312, opened 2021-01-27
Updates package.json to use arches as a dependency

chiatt:
Updates package.json to use arches as a dependency


chrabyrd:
works as expected!

Issue #314, opened 2021-01-29
object-search-step CSS fixes


chrabyrd:
accessory ticket to https://github.com/archesproject/arches/pull/7061 , adding in small CSS changes object-search-step

PR #315, opened 2021-01-29
small changes to better bound elements in tabbed workflow #314

chrabyrd:


chrabyrd:
merged into latest PR

Issue #302, opened 2021-01-13
Remove unnecessary workflow steps


chrabyrd:
`insert-prev-resource-as-related-resource-step` is currently unused. Should we remove this?

chiatt:
@chrabyrd Yeah, I think we should.

PR #320, opened 2021-02-02
Remove unused workflow step, #302

njkim:
Remove a unused workflow step, #302
insert-prev-resource-as-related-resource-step.js


chiatt (approved):
Meets requirements

PR #321, opened 2021-02-03
Update the physical thing workflow wastebin, #259

njkim:
Update definition of new physical thing workflow to avoid duplicates in the remove list. #259


chiatt (approved):
Works as expected

Issue #326, opened 2021-03-04
Calculate time span in the sample info workflow step


chiatt:
For the sampling date, users will enter a single date. This date corresponds to the `earliest start date` in the Sampling Activity Dates card.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/70c264d1-b6b8-4439-aa51-97e749935dd2)

The other dates will need to be calculated and saved to the tile.  These may change, but initially the `latest start date` can be the same as the earliest start date and the earliest end date and latest end date can be the end of the day.

njkim:
The sample taking workflow is being updated for the updated resource models.
#359

PR #334, opened 2021-03-29
Calculate a sampling time spans in workflow, #326

njkim:
Calculate the time spans in sampling workflow automatically, #326


chiatt:
This feature is no longer required

Issue #335, opened 2021-03-29
Thematic report template should have the ability to fetch data on load


chrabyrd:
twin ticket of: https://github.com/archesproject/arches/issues/7262

PR #337, opened 2021-04-01
Thematic report template fetches resource data on load  #335

chrabyrd:
twin pairing of https://github.com/archesproject/arches/pull/7274

solves #335


chiatt (approved):
Works as expected

PR #343, opened 2021-04-21
remove manifest function from digital resource model

annabelleee:
remove manifest function from digital resource model


chiatt:
Resolved in #344

PR #348, opened 2021-04-27
generic widget component thematic report template

chrabyrd:
To test:

1. Navigate to designer and set `Physical Thing`'s template to `Thematic Report`.
2. Create a new `Physical Thing` resource instance.
3. Navigate to either report or resource view
4. Interact with the `Names/Classification` tab. 
5. The other tabs do not have seed data, though they should handle it if desired.

The `hide empty report sections` toggle is functional. 
All other buttons are not.
There is some minor development silliness until we replace it with real logic ( `hasPlusSign` )


chiatt (approved):
Looks like a great start to thematic report components

Issue #353, opened 2021-04-30
UI issues in Add Things To Your Set step in Project Workflow


chiatt:
In the main panel on the left, each element in the search results should have padding around it and the description text should be visible in the body of the element. 

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/3a9b29d3-2893-4403-9589-2dd83c96a58f)

PR #357, opened 2021-05-04
Update css in the new project, #353

njkim:
Update the css in the new project workflow - add things step, #353 
It is updating the physical thing list


chiatt (approved):
Resolves #353

Issue #289, opened 2020-12-10
Create a step in the upload dataset workflow to upload candidate files for the dataset


chiatt:
Create a step in the upload dataset workflow to upload candidate files for the dataset. This step corresponds with step 2 in the mockups

chiatt:
Resolved in #409

PR #366, opened 2021-05-14
Updates to Upload Dataset workflow

chiatt:
re #289

- Adds initial instrument step,
- Move select phys thing step outside of upload dataset directory to be used in other workflows,
- Save selected phys thing name to local storage in select phys thing step.


njkim (approved):
Works as expected

Issue #358, opened 2021-05-04
Object search step not reloading selected objects 


chiatt:
Selected physical things do not appear in object search step if a user completes the object search step, advances, and then returns back to object search step.

PR #377, opened 2021-06-18
Fix add things step new project #358

njkim:
Fix the issue with repopulating the added list when reloaded by back button, #358
Also fix
1. the issue with ui (selection css persisted through the navigation between tabs)
2. the issue with adding a new one after after navigating away from the tab.


chiatt (approved):
Fixes #358

PR #385, opened 2021-06-24
Final step - Create project workflow, #381

njkim:
Update the final step of the create project workflow, #381


chiatt (approved):
Works as expected

PR #390, opened 2021-06-25
Fix create Project Add things step, #358

njkim:
Add related resource record on phys thing from Collection, #358 
To resolve the issue with only creating rr record on collection from things.


chiatt (approved):
Fixes #358

PR #400, opened 2021-07-08
Update analysis areas regions step to be compatible with core iiif-annotation reversion

chrabyrd:


chiatt (approved):
Fixes issue with newly created feature not appearing in features list.

PR #405, opened 2021-07-09
adds sample taking workflow image step

chrabyrd:


chiatt (commented):
@chrabyrd This looks good - it creates a manifest and dataset as expected, but I noticed it doesn't seem to create a tile on the parent physical thing establishing the relationship with the digital resource. I guess that can probably be added as part of the following step given that it will be needed to identify the preferred manifest.

PR #408, opened 2021-07-09
Detects analysis area modification

### LINKED ISSUE #399: 'Next' button in region step remains disabled after saving region
If a user creates a region in the region step of the analysis area workflow, the next button remains disabled and the user cannot advance.

aarongundel:
marks step as dirty, saves it so we can move forward to summary

fixes #399

### DIFF
--- a/afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
+++ b/afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
@@ -14,7 +14,7 @@ define([
 
         var objectStepData = params.form.externalStepData['objectstep']['data'];
         this.physicalThingResourceId = koMapping.toJS(objectStepData['sample-object-resource-instance'][0][1]);
-
+        
         var digitalResourceServiceIdentifierContentNodeId = '56f8e9bd-ca7c-11e9-b578-a4d18cec433a';
         var imageStepData = params.form.externalStepData['imagestep']['data']['image-service-instance'][0];
         this.manifestUrl = ko.observable(imageStepData.data[digitalResourceServiceIdentifierContentNodeId]);
@@ -31,14 +31,14 @@ define([
         this.hasExternalCardData = ko.observable(false);
 
         this.analysisAreaInstances = ko.observableArray();
-
+        
         this.selectedAnalysisAreaInstance = ko.observable();
         this.selectedAnalysisAreaInstance.subscribe(function(selectedAnalysisAreaInstance) {
             self.highlightAnnotation();
 
             if (selectedAnalysisAreaInstance) {
-                /* TODO: switchCanvas logic */
-
+                /* TODO: switchCanvas logic */ 
+                
                 self.tile = selectedAnalysisAreaInstance;
                 params.tile = selectedAnalysisAreaInstance;
                 self.physicalThingPartIdentifierAssignmentTile(selectedAnalysisAreaInstance);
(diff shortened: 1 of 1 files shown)


afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
```diff
@@ -174,11 +176,15 @@ define([
...
-        //     self.tile.save().then(function(data) {
-        //         self.observationInstances(self.card.tiles())
-        //     });
-        // };
+        this.saveAnalysisArea = function () {
+            params.form.complete(false);
+            params.form.saving(true);
+            self.analysisAreaInstances().forEach(
+                instance => params.form.savedData.push({ "data": instance.data })
+            );
```
chrabyrd, line 184: nit: this can just be `params.form.savedData(self.analysisAreaInstances());`

afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
```diff
@@ -174,11 +176,15 @@ define([
             self.selectAnalysisAreaInstance(newTile);
         };
 
-        // this.saveTile = function() {
-        //     self.tile.save().then(function(data) {
-        //         self.observationInstances(self.card.tiles())
-        //     });
-        // };
+        this.saveAnalysisArea = function () {
```
chrabyrd, line 179: nit: can this be named something like `saveWorkflowStep`?

afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
```diff
@@ -162,8 +163,9 @@ define([
             physicalThingAnnotationNode.active(true); 
         };
 
-        this.saveAnalysisAreaTile = function() {
-            self.tile.save().then(function(data) {
+        this.saveAnalysisAreaTile = function () {
+            self.tile.save().then(function (data) {
```
chrabyrd, line 167: nit throughout: can the spacing between `function` and arg be removed?

chrabyrd (commented):
Overall looks good! Just a few nits and it should be g2g

afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
aarongundel: Yes - it's also good to note that in my next PR, some of this is auto-formatted.  Perhaps some common style guide/formatter might be in order (prettier?)

afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
aarongundel: Makes much more sense after today

afs/media/js/views/components/workflows/analysis-areas-annotation-step.js
aarongundel: @chrabyrd this actually doesn't work; it throws an exception.

aarongundel: (error converting circular structure to JSON)

chrabyrd: ah actually I got it wrong.

let's map the `data` attr from each `analysisAreaInstance` into an array.

Then, have `params.form.savedData` consume the mapped array.

That way we're only saving the data we care about ( `analysisAreaInstance.data`), we're avoiding the JSON error, and we're preventing unnecessary observer/subscription signals by only interacting with `params.form.savedData` once.

chrabyrd (approved):
overall lgtm!

Issue #410, opened 2021-07-12
Summary step fails if 'Project Statement' not completed in second step.


chiatt:
If a user does not enter any text in the Project Statement's text area input (second step of the project workflow), the summary page will throw an error. To work around this user would have to return to the Project Statement and add a string in order to complete the workflow.

chiatt:
Closed in #411

PR #411, opened 2021-07-13
Fix the final step of create project workflow, #410

njkim:
Fix the issue of cause an error when statement is not available by mapping statements (1:n) and find a proper one, #410


chiatt (approved):
Works as expected

Issue #145, opened 2020-05-18
Physical thing search - suggest only terms that will return search results


chiatt:
Physical thing search in the Collection workflow shows terms from resource instances that are unrelated to physical things. That means that you can select terms that return no results because the term does not belong to a physical thing. The example below shows a dropdown in which users can select either Person or Physical thing instances. Only Physical Things should be available

![image](https://user-images.githubusercontent.com/1877663/125358477-cabef100-e31d-11eb-9ca0-f2589e17de4a.png)

PR #412, opened 2021-07-13
Filter the results in physical things step, #145

njkim:
Filter the dropdown selection in create project workflow to limit the selection to physical things, #145


chiatt (approved):
Works as expected

PR #413, opened 2021-07-13
Fixes physical thing analysis areas regions step save logic

chrabyrd:
Now on tile save, a new physical thing is either created or correctly edited, along with the reference tiles on the parent physical thing.


chiatt (approved):
Resolves 407

Issue #416, opened 2021-07-13
The add things step creates multiple collection resources


njkim:
The add things step in the create project workflow creates multiple collection resources

PR #418, opened 2021-07-14
Create set resource when the tile saved, #416

njkim:
To fix the issue of creating multiple collection/set resources in the add things step, #416


chiatt (approved):
Prevents multiple collections from getting created. re #416

PR #421, opened 2021-07-14
Upgrade postgres to v13.0 in Docker

chiatt:
Upgrade postgres to v13.0 in Docker

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,7 +49,7 @@ services:
 
     db_afs:
       container_name: db_afs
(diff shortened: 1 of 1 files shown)


aarongundel:
Looks good to me @chiatt - but it's worth noting that unsuspecting users who "docker compose up" after pulling this change are going to encounter fire and need to drop their afs stack and reload it.  Does anyone need to be notified of that?

aarongundel (approved):
Looks good to me.

PR #424, opened 2021-07-15
returns multiple thing objects when there are multiple sets in a project

### LINKED ISSUE #417: Select Physical Thing Step only allows users to select things from one collection
If a selected project has multiple collections, the select physical thing step will provide physical things from the first collection.

aarongundel:
Returns multiple thing objects when there are multiple sets/collections in a project

resolves #417


chiatt (approved):
Resolves #417

Issue #423, opened 2021-07-15
Lock steps in the upload dataset workflow


chiatt:
Completion of the second step should lock the first step. Completion of the third step should lock the second step.

PR #427, opened 2021-07-16
Lock first two steps of upload dataset workflow

chiatt:
Lock the each of the first two steps after their subsequent steps are saved. Depends on #409, re #423


njkim:
Works as described

Issue #439, opened 2021-07-19
Convert first four steps of the project workflow to component based step


chiatt:
The following steps need to be converted to the component based step:
```
Project Name
Project Statement
Project Timespan
Project Team
```

Project Timespan has hidden node and may require changes in this PR: https://github.com/archesproject/arches/pull/7403/files

PR #441, opened 2021-07-21
The create project workflow updated, #439 

njkim:
The create project workflow updated based on the component based steps, #439 

The hiding the nodes is dependent on 
https://github.com/archesproject/arches/pull/7403


chiatt (approved):
Resolves #439

PR #443, opened 2021-07-21
locks the object and image step in analysis area workflows

### LINKED ISSUE #432: Lock steps in the analysis areas workflow
Saving the second step should lock the first and saving the third should lock the second

aarongundel:
locks the object and image step in the analysis area workflow.  Needs PR [7620](https://github.com/archesproject/arches/issues/7620) for core.  Also needs a fix for dirty tiles to validate image step locking, pending discussion.  

closes #432

### DIFF
--- a/afs/media/js/views/components/plugins/analysis-areas-workflow.js
+++ b/afs/media/js/views/components/plugins/analysis-areas-workflow.js
@@ -13,6 +13,7 @@ define([
             var self = this;
 
             this.resourceId = ko.observable();
+            
 
             params.steps = [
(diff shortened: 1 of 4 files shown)


chiatt (changes requested):
This looks good. Step one is disabled and I see the radio button is disabled in the image step, but it is still possible to create a new manifest which will automatically change the radio button to the newly created manifest. To avoid this, I think we will need to disable the 'Create New Image Service' button as well.

chiatt (approved):

PR #445, opened 2021-07-21
Fixes a regression with the analysis area annotation step

### LINKED ISSUE #444: Regression in the analysis areas annotation step advancement
Step never shows as dirty after creating annotations.  This is a regression, originally solved for #399 - line got knocked out when undoing some autoformatting code.

aarongundel:
There's a regression here; I dropped a line when fixing some auto-formatting code.  This PR adds that line (marking the step dirty) back whenever there's a save action.

fixes #444


chrabyrd (approved):
works as expected

PR #452, opened 2021-07-22
Fix for dockerized celery not starting

### LINKED ISSUE #451: dockerized celery doesn't start
The dockerized celery worker isn't starting properly.  It currently needs to be manually started.  It should come up automatically when the afs container starts

aarongundel:
Fix for dockerized celery not starting; after this it should come up automatically

fixes #451

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ RUN set -ex \
         python3.8-distutils \
         python3.8-venv \
         dos2unix \
+        supervisor \
     " \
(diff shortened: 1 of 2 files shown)


chiatt:
The issue here is that this causes resource instances to load in the background using celery. I think it would be better if we could control that in order to better monitor/debug the load, so I'd like archesproject/arches#7628 to be resolved before merging this in.

aarongundel:
no longer relevant

PR #457, opened 2021-07-23
saves statement and name type values

### LINKED ISSUE #454: Save statement and name type values in instrument info step
The instrument info step saves a tile for Name and a tile for Statement. A corresponding type should be saved with each tile.

aarongundel:
saves statement and name type values for nodes created in the instrument info step

closes #454


chiatt (approved):
Resolves #454

PR #461, opened 2021-07-23
Fix missing relationship between project and collection in Project workflow

chiatt:
re #460


aarongundel:
works as advertised!

aarongundel (approved):

PR #469, opened 2021-07-27
Validate the physical thing in upload dataset workflow

### LINKED ISSUE #463: Ensure selected phy thing in first step of upload dataset is valid
In order to proceed from the first step, the selected physical thing must have:

1. A digital resource that references a manifest
2. One or more 'parts' associated with a manifest

If the selected physical thing does not meet these requirements, the user should be informed that they need to complete the analysis area or sample taking workflow before proceeding with the upload dataset workflow.

aarongundel:
Validate the physical thing for digital references with manifest types and parts with manifests.

Stickler will definitely not like the optional chaining.

resolves #463


chiatt (approved):
Resolves #463

Issue #470, opened 2021-07-27
Upload dataset sample location step fails to load


aarongundel:
There is an issue with loading a previous step's state.  Looks like one of the recent merges changed the name of the state's data object.

PR #471, opened 2021-07-27
bugfix for the sample location step

aarongundel:
Changes unique instance name back to instrument-info, where the sample location step is looking.


chiatt (approved):
Resolves #470

Issue #474, opened 2021-07-28
Dataset ids should be saved to local storage in select-dataset-files-step


chiatt:
The Dataset instances created in the select-dataset-files-step need to be saved to local storage so that they are available to the following dataset interpretation step.

PR #475, opened 2021-07-28
dataset ids should be saved

aarongundel:
This saves newly generated dataset ids in the select dataset files step.  Also reduces some nesting.

@chiatt even though I validated that this does indeed save the ids, it looks insufficient on its own.  The ids will need to be used in the subsequent step, and it looks like this (dataset files) step doesn't load its state from local storage properly, so clears out every time it's saved.  These seem like separate tickets, though.


chiatt (approved):
Resolves #474

PR #476, opened 2021-07-28
Adds alerts for required nodes when creating an annotation

### LINKED ISSUE #473: Alert user when saving an analysis area without adding a required node
In the regions step of the Analysis Area workflow, name is a required node, but if a user does not fill it out and tries to save an analysis area, an error on the backend will cause the UI to hang.

If a user fails to save without the required node, they should get the alert banner instructing them to complete the form.

aarongundel:
Adds alerting when name and geometry annotation are not provided

resolves #473


chiatt (approved):
Resolves #473

chiatt (approved):
Resolves #473

PR #481, opened 2021-07-30
adds accordion functionality to the management of datasets/uploads

### LINKED ISSUE #467: Change list of parts in upload dataset step to accordian
Currently the upload dataset step allows the user to select a part and then associate files with the selected part. This creates some ambiguity when the user selects a different part and then is able to associate the same files with the newly selected part that were associated with the first part.

To address this issue, a part in the parts list should be able to expand to show the files associated with it directly under the name of the part. The input for the dataset name should also appear under the name of the part. In the list of uploaded files, only those files that remain unassigned to a part should appear in the list.

aarongundel:
resolves #467 

Adds accordion functionality for file association with datasets.

### DIFF
--- a/afs/media/css/project.css
+++ b/afs/media/css/project.css
@@ -460,6 +460,50 @@
     padding: 0px;
 }
 
+.upload-dataset-select-dataset-files {
+    margin-top: 20px;
+}
+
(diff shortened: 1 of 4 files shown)


chiatt (approved):
Resolves #467

PR #491, opened 2021-08-03
remove unnecessary subscriber, reset step value on thing change

### LINKED ISSUE #490: Clear phys thing when project changes in select-phys-thing-step
If a user selects a project in the select-phys-thing-step, then selects an associated physical thing and then changes the project, the physical thing initially selected persists. The problem is that the step assumes since the project and physical thing are both selected, the step is complete and the user can proceed. This can allow a user to select a physical thing that does not actually belong to the selected project.

To prevent this when a user changes the project, the physical thing input should be set back to `undefined`.

aarongundel:
remove unnecessary subscriber, reset step value on thing change

resolves #490


chiatt (approved):
Works as expected

Issue #367, opened 2021-05-19
Create dataset interpretation component


chiatt:
The upload dataset and review dataset workflows need a component that allows a user to chart and edit files from multiple digital resource instances. This component should occupy a layout section in the component based step. See pages 24 and 30 in the mockup: https://www.dropbox.com/home/Arches4/Mockups/AfS?preview=DISCO+Workflows+v10.1.pdf

PR #496, opened 2021-08-03
file intrepretation step

apeters:
re #367


chiatt (approved):
Allows user to view chart and save interpretation statements

PR #498, opened 2021-08-03
adds parameter to select-phys-thing-step to validate the physical thing

### LINKED ISSUE #494: Physical thing selector shouldn't validate the thing in the sample taking workflow
The physical thing selector step shouldn't validate the thing in the sample taking workflow, since it's used to set it up.  Probably solved easily enough by passing a config parameter to validate.

aarongundel:
adds parameter to select-phys-thing-step to validate the physical thingng (in the case of upload dataset) - defaults to false

resolves #494


chiatt (approved):
Resolves #494

PR #501, opened 2021-08-04
If physical thing validation is turned off, don't check for a manifest

### LINKED ISSUE #500: Skip dataset/manifest validation in select-phys-thing-step in Sample Taking
A dataset/manifest gets created during the course of the Sample Taking workflow, so that does not need to be included in the select-phys-thing-step validation.

aarongundel:
Avoids manifest check if physical thing validation is turned off. 
resolves #500


chiatt (approved):
Resolves #500

PR #505, opened 2021-08-04
edit link visibility fix

### LINKED ISSUE #495: Edit link missing in search results
The edit link is unavailable in the search results list. Note that AfS has a custom `search-results` search component.

aarongundel:
uses a filter on the user object to toggle the visibility of the edit link depending on permissions
fixes #495


chiatt (commented):
Thanks @aarongundel this fixes the issue, but it is the old way that we were checking for permissions which resulted in a lot of SQL queries. `user_can_edit` is the new approach, but I can see now that `user_can_edit` is not available to the search component. I've come up with a way in core arches to pass the permission status down to the component using JS, so I'm going to close this PR.

PR #510, opened 2021-08-05
changes for core ticket 7655

chrabyrd:
WARNING! NOT BACKWARDS-COMPATIBLE!

these update the workflow config files to be compatible with new step path logic


chiatt (approved):
Overall everything works as expected. There is an error in the final step of the new project workflow because the 'resourceid' is no longer available, but that probably is due to merging in the latest from master and can be addressed in another ticket.

Issue #503, opened 2021-08-04
File Interpretation Step Loads only one dataset


chiatt:
The file Interpretation Step Loads only one dataset created in the proceeding step even when multiple datasets should be available.

Issue #504, opened 2021-08-04
File Interpretation Step does not allow user to proceed to summary


chiatt:
After a user saves an interpretation for at least one dataset, the step should enable the next button and allow the user to proceed to the summary.

Issue #502, opened 2021-08-04
File Interpretation step needs workflowstepclass to remove padding


chiatt:
Padding around the File Interpretation step should be removed using a worflowstepclass. It may be possible to use the same workflowstepclass used in the proceeding Sample Location step. In that case the class should probably be renamed to be more generic.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/ac8c3d41-b572-4bd6-8278-1c6b9f0a0413)

PR #511, opened 2021-08-05
allow multiple datasets to load re #503,  add workflowstepclass

apeters:
re #502, allow user to go to the next step re #504


chiatt (approved):
Resolves #503, #502, #504

Issue #515, opened 2021-08-06
Remove dependency on search-result.htm override in new project workflow


chiatt:
Remove dependency on search-result.htm override in new project workflow

PR #516, opened 2021-08-06
Remove need for search-result.htm override 

chiatt:
Removes the need to override the search-result search component as used in the New Project workflow. re #515 
**Requires PR** https://github.com/archesproject/arches/pull/7690


chrabyrd (approved):
lgtm!

PR #526, opened 2021-08-12
dataset cancel/save button fix

### LINKED ISSUE #506: Select dataset files step's save/cancel buttons remain enabled regardless of dirty state
Enabled state of buttons need to reflect the dirty state

aarongundel:
Fixes dirty cancel/save buttons in the upload dataset step.
Resolves #506


chiatt (approved):
Works as expected

PR #529, opened 2021-08-13
Update add things step to a component based step, re #430

njkim:
Update add-things-step in create-workflow to a component based step, re #430


chrabyrd (approved):
lgtm!

Issue #488, opened 2021-08-02
Remove cruft workflow step files


chiatt:
There are workflow files that are no longer needed in `/views/components/workflows`. These need to be identified and removed.

PR #531, opened 2021-08-16
removes afs workflow cruft #488

chrabyrd:
solves:

- https://github.com/archesproject/arches-for-science-prj/issues/488
- https://github.com/archesproject/arches-for-science-prj/issues/487


chiatt (approved):
Resolves #488

PR #538, opened 2021-08-17
Upload dataset final step, #387

njkim:
Update the final step of the upload dataset workflow, re #387


chiatt (approved):
Works as expected

PR #539, opened 2021-08-17
 Add annotations to upload dataset final step, #387

njkim:
Add annotations to the final step of upload dataset workflow, re #387


chiatt (approved):
Works as expected

PR #540, opened 2021-08-17
define resourceid and comment out annotation component

chiatt:


chiatt:
No longer an issue in latest

Issue #541, opened 2021-08-17
update sample-taking-summary-step for demo


chrabyrd:

chiatt:
Fixed in #542

PR #542, opened 2021-08-17
541 cbyrd updates sample taking summary step for demo

chrabyrd:
solves #541


chiatt (approved):
Gets final step working again

PR #544, opened 2021-08-17
updates cantaloupe to latest

### LINKED ISSUE #543: cantaloupe latest tag removed; switch to versioned
The _latest_ tag was removed from the ucla cantaloupe image; as a result docker-compose up no longer works.  Probably for the best we don't use it anyway and switch to a version based tag.

aarongundel:
resolves #543 
adds version tag to cantaloupe for docker


chiatt (approved):
Works as expected

Issue #548, opened 2021-08-18
New project workflow fails to create relationship to collection


chiatt:
When physical things are added to a collection, the relationship between the collection and the physical thing is created, but not in the other direction. The physical thing does not have a 'Used in Collection' tile. This causes other workflows that rely on this relationship to fail.

PR #549, opened 2021-08-18
Allow saveOnQuit to be called from add-things-step

chiatt:
Allow saveOnQuit to be called from add-things-step in order to create relationship between a physical thing and a collection, re #548 and archesproject/arches#7719

QA using https://github.com/archesproject/arches/pull/7720


njkim (approved):
Works as expected

Issue #462, opened 2021-07-24
Update the summary steps to use the beta json endpont


chiatt:
Workflow summary steps should leverage the beta version of the json endpoint:
`resources/[resourceid]?format=json&indent=3&compact=false&v=beta`

PR #550, opened 2021-08-18
Update the final steps to use beta endpoint, #462

njkim:
Update to final steps to use disambiguated end point version beta, #462


chiatt (approved):
Works as expected

PR #554, opened 2021-08-20
add-things-step replaces research-collection-step, re #488

chiatt:
@njkim replaced research collection step with the add-things-step, so the research collection step is no longer needed. re #488


njkim (approved):
works as expected

Issue #547, opened 2021-08-17
Add links to analysis area physical things in summary step


chiatt:
In the summary step there is a link to the parent physical thing and the datasets, but the analysis area physical things also need to be listed.

PR #555, opened 2021-08-20
Update the final step of Analysis Areas, #547

njkim:
Add the list & link of regions in the final step of Analysis Areas, #547


chiatt (approved):
Works as expected

Issue #551, opened 2021-08-18
Sample location step hangs if data is incomplete


chiatt:
In the sample location step if a user fails to create a feature for a sample location before saving, the page will display the loading indicator but will never finish saving due to a required node error. Instead, the Alert banner should drop notifying the user which nodes need to be completed to save a sample location.

PR #556, opened 2021-08-21
Error handling for missing sample loc & name, #551

njkim:
Add an alert and exit the saving chain if the sample location or the sample name is missing, #551


chrabyrd (approved):
looks good

Issue #563, opened 2021-08-23
Final step of new project workflow fails due to missing resourceid


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/e7ba6950-421f-414c-abcb-767322ff4be3)

chiatt:
Fixed in latest

PR #566, opened 2021-08-24
Fix the create project's final step, #563

njkim:
Fix the issue of missing resourceid in the final step of create-project workflow, #563 
By adding `externalStepData` and reintroducing the `shouldtrackresource`.


chiatt:
No longer needed in latest due to changes in how external step data is managed.

Issue #577, opened 2021-08-27
Alert messages not appearing in workflows


chiatt:
There is an error message that should fire when a user tries to save a dataset without a name on the upload data set workflow. The error gracefully prevents the user from saving, but the Alert banner explaining why the user cannot save does not appear.

chiatt:
Fixed in #579

PR #579, opened 2021-08-27
Update alerts in workflows, #577

njkim:
Update the alerts in the workflows, #577
Replacing `params.form.alert` with `params.pageVm.alert`


chiatt (approved):
Alerts work as expected

chiatt (approved):
Works for me

Issue #583, opened 2021-08-31
Add motivation statement to the sample info step


chiatt:
Currently the motivation for sampling is found in the sample location step. This input should be added to the Sample Info step. re archesproject/arches-for-science#581

chiatt:
Resolved in #584

PR #584, opened 2021-08-31
Add sampling motivation, #583 

njkim:
Add sampling motivation for the while sampling activity in the sample info step, #583


chiatt (approved):
Solves #583

Issue #580, opened 2021-08-30
Sample for sampling region should be automatically named 


chiatt:
When a user creates a new sample location a physical thing for the region as well as a physical thing for the sample itself are created. These currently are given the same name, however the a sampling region should automatically be given a distinct name.

chiatt:
Resolved in #586

Issue #581, opened 2021-08-31
Make sample motivation in sample location step a statement on the physical thing


chiatt:
Currently the 'motivation for sample' in the sample location step updates the motivation node on the sampling activity. Instead it should update a statement on the physical thing much like the 'sample description'. The type for this tile will be 'motivation'. This concept value is not yet available, so a placeholder concept should be used until 'motivation' is added.

chiatt:
Resolved in #586

PR #586, opened 2021-09-01
Update sample location step, #580, #581

njkim:
Update the sample location step of sample taking workflows, #580, #581


chiatt (approved):
Resolves #580 and #581

Issue #585, opened 2021-08-31
Add a loading spinner when manifest datasets are loading in Image step


chiatt:
A loading spinner is needed while datasets are loading in image step so that the user doesn't presume that a new manifest needs to be created when one already exists, but is not yet displayed.

chiatt:
This was solved but has regressed in the latest.

chiatt:
This should apply to the image step in both the analysis area workflow as well as the sample taking workflow.

PR #590, opened 2021-09-02
Loading spinner while loading saved manifest, #585

njkim:
Add loading while saved manifest is loading, #585


chiatt (approved):
Resolves #585

Issue #587, opened 2021-09-01
Ensure previously created datasets load in the upload dataset step


chiatt:
If a user creates an anaysis area (Area A), then goes through the upload dataset workflow, they will create a new dataset related to Area A. If they go through the analysis area workflow a second time, create a second area (Area B), and then execute the upload dataset workflow, they will not see the dataset for Area A. This may create confusion because they will see both Area A and Area B but think that neither of them have any associated files. 
To address this issue the upload dataset workflow should show each analysis area along with those datasets that have already been created for existing areas.

chiatt:
Fixed in #591

PR #591, opened 2021-09-02
Load existing datasets in upload dataset workflow

chiatt:
Loads existing datasets for all analysis areas in the upload dataset step. re #587


njkim:
works as described

Issue #582, opened 2021-08-31
Area name templating


chiatt:
Currently users can name analysis areas anything that they like. Instead we need to ensure the at the name of the parent physical thing is included in the name. The users input should follow the physical thing name in brackets, and the complete name to be saved should be previewed below the name input.

chiatt:
duplicate

chiatt:
Resolved in #595

PR #595, opened 2021-09-03
Automatically create an area name, #582

njkim:
Automatically create an area name which include the parent physical thing name, #582


chiatt (approved):
Resolves #582

PR #598, opened 2021-09-03
fixes docker volumes not working for postgres data

aarongundel:
fixes docker volumes not storing postgres data.  Temporary fix until shared containers are in use.


chiatt (approved):
Works for me

Issue #599, opened 2021-09-03
Missing samples in the final step.


njkim:
Sample location list in the final step summary are missing samples if the sample does not have statements (description, motivation) at all.

PR #600, opened 2021-09-03
Missing samples final summary, #599

njkim:
Fix the issue of missing samples when the sample does not have any statement at all, #599
Also, add the link between the annotation feature and the list in the table.


chiatt (approved):
Works as expected

Issue #520, opened 2021-08-10
Prevent analysis areas from appearing in sample area list


chiatt:
Analysis areas should not appear in the list of sample locations in the sample location workflow step. Likewise, sample locations should not appear in the list of analysis areas in the Regions step of the Analysis Areas workflow. Currently there is not a simple way to distinguish these two kinds of parts/physical things. Having a node on the part branch to define the type of part would help with the distinction.

PR #601, opened 2021-09-07
Filter analysis areas in sample location step

chiatt:
Filter analysis areas from the list of sample locations in the sample taking location step. re #520


chiatt:
To QA, go through the analysis areas workflow and create an analysis area. Open the newly created analysis area in the editor and set the Type in the Classification card to `ceramic`. Once that's done, go through the Sample Taking workflow and select the parent of your analysis area as your Physical Thing in the first step. When you get to the sample location step, your analysis area should not appear in the list of sample locations.

chiatt:
Closing in lieu of #645

PR #602, opened 2021-09-07
new project workflow saving

### LINKED ISSUE #597: Final step in new project workflow is not saving relationship to collection
When a user completes the new project workflow and presses 'save and complete workflow' button, a tile is saved to each physical thing creating a relationship to the project's collection. This tile is no longer saving. To fix this issue, instead of saving the relationship at the end of the workflow, we should save it when a user saves the 'Add Objects to Your Project' step.

aarongundel:
The problem is that the step is overriding params.form.saveOnQuit - that saves at the end of the workflow (Save and Complete Workflow).  To save at the end of the step, we need to call that method within the step's save method.  

Fixes #597


chiatt (approved):
Fixes #597

PR #604, opened 2021-09-07
standardize workflow save buttons

aarongundel:
standardizes workflow save buttons; removes save & undo from file interpretation, sample location, and select files step


chiatt (approved):
Works as expected

Issue #613, opened 2021-09-09
Cancel button not working for select-dataset-files-step


chiatt:

chiatt:
Fixed in #615

PR #615, opened 2021-09-10
select dataset files cancel

aarongundel:
Adds functionality to cancel button in select dataset files step.  

Finishes #613


chiatt (approved):
Looks good to me

PR #617, opened 2021-09-10
Add setting to no longer ignore tile cardinlity constraint, archespro…

ryan86:
…ject/arches#7781

Add `BYPASS_CARDINALITY_TILE_VALIDATION = False` to settings.py to no longer bypass cardinality check when saving tiles.

### DIFF
--- a/afs/settings.py
+++ b/afs/settings.py
@@ -2,7 +2,6 @@
 Django settings for afs project.
 """
 
-from arches.arches.settings import BYPASS_CARDINALITY_TILE_VALIDATION
 import os
 import arches
 import inspect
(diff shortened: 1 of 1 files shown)


afs/settings.py
```diff
@@ -2,6 +2,7 @@
 Django settings for afs project.
 """
 
+from arches.arches.settings import BYPASS_CARDINALITY_TILE_VALIDATION
```
chiatt, line 5: @ryan86 Can you remove this import? It's actually already covered by the `import *` on line 10.

afs/settings.py
```diff
@@ -152,7 +155,7 @@
 USER_GRAPH_PERMITTED_CARDS_TIMEOUT = 3600 * 24 * 30  # seconds * hours * days = ~1mo
 USER_GRAPH_CARDWIDGETS_TIMEOUT = 3600 * 24 * 30  # seconds * hours * days = ~1mo
 
-MOBILE_OAUTH_CLIENT_ID = ""  #'9JCibwrWQ4hwuGn5fu2u1oRZSs9V6gK8Vu8hpRC4'
+MOBILE_OAUTH_CLIENT_ID = "7celI8OVqhC5UViw3UZL6pbCwLuxIbiOuVfn4YGu"  #'9JCibwrWQ4hwuGn5fu2u1oRZSs9V6gK8Vu8hpRC4'
```
chiatt, line 158: @ryan86 Look like this change snuck in. Can you change back to an empty string?

chiatt (changes requested):
Thanks @ryan86, this fixes the issue but there are just a couple minor changes to make.

chiatt (approved):

Issue #616, opened 2021-09-10
Create Project does not delete collection created when delete and quit clicked


njkim:
Create Project Workflow creates a project resource instances and a collection resource instance along with relationship between resources and physical resource instances.

When a user click the delete workflow, it only delete the project resource not the collection resource.

chiatt:
Nice catch @njkim

chiatt:
Fixed in latest

Issue #610, opened 2021-09-08
Allow users to delete things from a collection


chiatt:
In the add-things-to-step workflow step, users can add items to a collection, but if they remove them after they are saved, this change does not persist in the database. Users should be able to add and remove items from a collection even after the relationship tile is saved.

chiatt:
Resolves in #618

PR #618, opened 2021-09-10
Manage the added things in create-project workflow, #610

njkim:
Add the capability to manage (add and remove) the added physical things in create project workflows, #610 
Also, resolve the issue with not removing the collection resource when the workflow is deleted, #616

### DIFF
--- a/afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
+++ b/afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
@@ -174,27 +174,41 @@ define([
                 self.removedValues(self.startValue().filter(val => !self.value().includes(val)));
                 self.addedValues(self.value().filter(val => !self.startValue().includes(val)));
 
-                Promise.all(saveCollectionRelationships()).then(
-                    Promise.all(removeCollectionRelationships()).then(
-                        createActivityUsedSet().then(function(){
-                            console.log("finished", ko.unwrap(self.value))
-                            self.savedData(
-                                {
-                                    value: ko.unwrap(self.value),
-                                    projectResourceId: ko.unwrap(self.projectResourceId),
-                                    collectionResourceId: ko.unwrap(self.collectionResourceId),
-                                    collectionTileId: ko.unwrap(self.collectionTileId),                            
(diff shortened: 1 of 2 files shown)


afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
```diff
@@ -225,59 +101,55 @@ define([
...
 
-        this.tile.subscribe(function(tile) {
-            self.startValue = tile.data[ko.unwrap(params.nodeid)]();
-            if (self.startValue) {
-                self.startValue.forEach(function(item) {
-                    self.value.push(koMapping.toJS(item));
-                });
+        this.initialize = function(){
+            if (params.value()) {
+                var cachedValue = ko.unwrap(params.value);
```
aarongundel, line 114: nit: avoid var

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
```diff
@@ -292,11 +164,54 @@ define([
...
+                    'transaction_id': params.form.workflowId
+                },
             }).done(function(data) {
                 self.collectionResourceId(data.resourceinstance_id);
+                self.collectionTileId(data.tileid);
+
+                self.removedValues(self.startValue().filter(val => !self.value().includes(val)));
+                self.addedValues(self.value().filter(val => !self.startValue().includes(val)));
+
+                createActivityUsedSet().then(
```
aarongundel, line 178: async/await could be used to reduce nesting (since Promise.all returns another promise, as does createActivityUsedSet)

aarongundel (changes requested):
Looks good and works as expected!  Few minor issues...

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
chiatt: I think this would be a nice refactor and definitely worth noting, but we can probably go ahead and merge and return to this later.

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
njkim: Thank you for catching this. updated.

Issue #620, opened 2021-09-14
Analysis Area Final Step fails when canvas is not available


njkim:
The final step appears to happen when the geom feature isn't available for the annotation thus canvas is not available.

PR #621, opened 2021-09-14
Update analysis area final step, #620

njkim:
Update the final step of analysis area to fix an error when annotation geometry is not available. #620 
Also, fix the issue with the not picking the right manifest when there are multiple preferred manifest are present.


chiatt (approved):
Works well for me

PR #622, opened 2021-09-14
Add loading spinner while manifest is loading, #585

njkim:
Add loading spinner to image-step in the sample taking and analysis area workflows, #585 
Improved to start the spinner while page is loading before create button is clickable (vs. after finish loading the page)


chiatt (approved):
Works well for me

Issue #623, opened 2021-09-14
Cancel button does not work for description and motivation in sample location step


chiatt:
User is unable to undo changes to the description and motivation values in the sample location step of the sample taking workflow.

chiatt:
Resolved in #624

PR #624, opened 2021-09-14
Reset descriptions when canceled, #623

njkim:
The values of sampling descriptions and motivations will be reset, when the cancel button is clicked, #623


chiatt (approved):
Looks good to me

Issue #625, opened 2021-09-14
Save the classification type when saving an analysis area


chiatt:
When an analysis area is saved in the analysis areas workflow, we need to save its type so that the sample taking step can recognize it as an analysis area.

Analysis area type: `19ef147a-3703-41b4-bfa3-d4c66e333b0b`
Classification Nodegroupid = `8ddfe3ab-b31d-11e9-aff0-a4d18cec433a`

PR #627, opened 2021-09-15
Save classification to tell area vs sample, #625

njkim:
Save the physical thing's classification tile after name is created hence resource_id is available, #625
Currently the classification is indicating "ceramic"


chiatt (approved):
Works as expected

Issue #639, opened 2021-09-17
Implement name scene in reports


njkim:
The remaining graphs to implement name scene are:

- basic components
- [x] Modification
- [x] Set 
- [x] Project
- add exact match
- [x] Group
- [x] Place
- [x] Period
- more complicated
- [ ] Person

PR #640, opened 2021-09-17
Implementing the name scene in reports w/o exact-match, #639

njkim:
Includes
 Modification
 Set
 Project
Observation


aarongundel (approved):
Not sure on impact of activity to project rename - checked with Namjun, and it seems that there's been a rename.  So approving based on that.

PR #641, opened 2021-09-17
Implementing the name scene in reports with exact-match, #639

njkim:
Includes
 Group
 Place
 Period

### DIFF
--- a/afs/media/js/views/components/reports/group.js
+++ b/afs/media/js/views/components/reports/group.js
@@ -1,9 +1,43 @@
-define(['jquery', 'underscore', 'knockout', 'arches', 'viewmodels/tabbed-report', 'utils/resource'], function($, _, ko, arches, TabbedReportViewModel, resourceUtils) {
+define([
+    'jquery',
+    'underscore',
+    'knockout',
+    'arches',
+    'utils/resource',
+    'utils/report',
+    'views/components/reports/scenes/name'
+], function($, _, ko, arches, resourceUtils, reportUtils) {
     return ko.components.register('group-report', {
         viewModel: function(params) {
             var self = this;
             params.configKeys = ['tabs', 'activeTabIndex'];
-            TabbedReportViewModel.apply(this, [params]);
+            Object.assign(self, reportUtils);
+            self.sections = [
+                {'id': 'name', 'title': 'Names and Classifications'}, 
+            ];
+            self.reportMetadata = ko.observable(params.report?.report_json);
+            self.resource = ko.observable(self.reportMetadata()?.resource);
+            self.displayname = ko.observable(ko.unwrap(self.reportMetadata)?.displayname);
+            self.activeSection = ko.observable('name');
+            self.dataConfig = {
+                exactMatch: 'exact match',
+            };
+            self.nameCards = {};
+
+            console.log(self.resource())
+
+            if(params.report.cards){
+                const cards = params.report.cards;
+                
+                self.cards = self.createCardDictionary(cards)
+
+                self.nameCards = {
+                    name: self.cards.name,
+                    identifier: self.cards.Identifier,
+                    exactMatch: self.cards.ExactMatch,
+                    type: self.cards.Classification
+                }
+            }
 
             if (params.summary) {
 
--- a/afs/media/js/views/components/reports/group.js
+++ b/afs/media/js/views/components/reports/group.js
@@ -24,8 +24,6 @@ define([
             };
             self.nameCards = {};
 
-            console.log(self.resource())
-
             if(params.report.cards){
                 const cards = params.report.cards;
                 
--- a/afs/media/js/views/components/reports/scenes/name.js
+++ b/afs/media/js/views/components/reports/scenes/name.js
@@ -110,12 +110,12 @@ define(['underscore', 'knockout', 'arches', 'utils/report','bindings/datatable']
                     }));
                 }
 
-                let exactMatchData = params.data()[self.dataConfig.exactMatch];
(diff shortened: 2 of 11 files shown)


afs/media/js/views/components/reports/group.js
```diff
@@ -1,9 +1,43 @@
...
+            self.reportMetadata = ko.observable(params.report?.report_json);
+            self.resource = ko.observable(self.reportMetadata()?.resource);
+            self.displayname = ko.observable(ko.unwrap(self.reportMetadata)?.displayname);
+            self.activeSection = ko.observable('name');
+            self.dataConfig = {
+                exactMatch: 'exact match',
+            };
+            self.nameCards = {};
+
+            console.log(self.resource())
```
aarongundel, line 27: console.log should be removed

afs/media/js/views/components/reports/scenes/name.js
```diff
@@ -110,9 +110,14 @@ define(['underscore', 'knockout', 'arches', 'utils/report','bindings/datatable']
...
                 }
 
-                if(self.dataConfig.exactMatch){
-                    self.exactMatch(self.getNodeValue(params.data(), self.dataConfig.exactMatch));
+                let exactMatchData = params.data()[self.dataConfig.exactMatch];
+                if(exactMatchData) {
+                    if(exactMatchData.length === undefined){
+                        exactMatchData = [exactMatchData]
+                    }
+                    self.exactMatch(exactMatchData.map(x => x["@display_value"]))
```
aarongundel, line 118: Instead of directly accessing @display_value here, it'd be better to call self.getNodeValue on the node itself - that way it can account for the display of undefined values and such.  so self.getNodeValue(x).  In addition, it's probably better to call self.getRawNodeValue instead of accessing self.dataConfig.exactMatch array style (L 113)

afs/templates/views/components/reports/scenes/name.htm
```diff
@@ -118,12 +118,11 @@ <h2 class="afs-report-section-title"><i data-bind="click: function() {toggleVisi
     <!-- Collapsible content -->
     <div data-bind="visible: visible.classifications" class="afs-report-collapsible-container pad-lft">
 
-        <!-- ko ifnot: identifiers().length -->
+        <!-- ko ifnot: type || exactMatch().length -->
         <div class="afs-nodata-note">No classification for this resource</div>
```
aarongundel, line 122: good catch above!  You'll also want to change this "no classification " to "no exact match"

afs/templates/views/components/reports/scenes/name.htm
```diff
@@ -146,7 +145,15 @@ <h4 class="afs-block-key">Type: </h4>
                         </div>
                         <div class="afs-block-attributes">
                             <h4 class="afs-block-key">Exact Match: </h4>
-                            <div class="afs-block-value"><a data-bind="if: exactMatch() != '--', attr: {href: exactMatch}"><span data-bind="text: exactMatch"></span></a><span data-bind="if: exactMatch() == '--'">--</span></div>
+                            <div class="afs-block-value">
+                                <!-- ko foreach: exactMatch -->
+                                <a data-bind="if: $data != '--', attr: {href: $data}"><span data-bind="text: $data"></span></a>
```
aarongundel, line 150: might want to wrap this anchor in a span and put the "if" there - IIRC the if check on the anchor will delete the inner content from it - but it might still style out with a width and be (weirdly) clickable.

afs/templates/views/components/reports/scenes/name.htm
```diff
@@ -146,7 +145,15 @@ <h4 class="afs-block-key">Type: </h4>
...
+                                <span>;</span>
```
aarongundel, line 151: Something like , might be better suited here so there's no trailing semicolon, or just toss that if onto the span itself.

aarongundel (changes requested):
Looks good, a few minor requests...

afs/media/js/views/components/reports/group.js
njkim: updated, thanks.

afs/media/js/views/components/reports/scenes/name.js
njkim: updated

afs/templates/views/components/reports/scenes/name.htm
njkim: I am removing this because both the type and the exact match seem to be showing "--" when there is not values?

afs/templates/views/components/reports/scenes/name.htm
njkim: updated

afs/templates/views/components/reports/scenes/name.htm
njkim: updated

afs/templates/views/components/reports/scenes/name.htm
aarongundel: pretty sure this is the same issue I keep seeing re: "dummy values" for these nodes.  That causes it to look like there's a record when there actually isn't.  In any case, removal is fine for now.

njkim: https://github.com/archesproject/arches-for-science-prj/blob/reporting_dev/afs/media/js/utils/report.js#L45
This line is useful when we have multiple values in a single line, (eg. name_content, name_type, etc) but if the node returning a single value (as in exact match or type) it can be trickier, I guess.

aarongundel (approved):
looks good!

Issue #642, opened 2021-09-17
Implement the description scene in the reports


njkim:
- [x] Digital Resource
- [x] Instrument
- [x] Physical Thing
- [x] Textual Work
- [x] Sampling Activity
- [x] Provenance Activity
- [x] Set
- [x] Project
- [x] Observation
- [x] Modification
- [x] Group
- [x] Place
- [x] Period
- [ ] Person
- [ ] Visual Work

PR #644, opened 2021-09-20
Description scene reports, #642

njkim:
The description is added to the reports. #642 

Includes:
    Provenance activity, textual work, sampling activity, collection, project, modification, observation, group, period, place
    
The remaining graphs are:
    Person, visual work


aarongundel (changes requested):
looks good!  no real concerns other than a few trailing console.logs

PR #645, opened 2021-09-20
Filter analysis areas and sample locations

chiatt:
Filters analysis areas from the list of sample locations in the sample taking location step, and filters sample locations from the analysis areas step and divides sample locations and analysis areas into two different annotation overlays. Requires archesproject/arches#7827 to QA. re #520 and #626


njkim (approved):
It works great.
The popup of non-editable feature is particularly the nice addition.

Issue #651, opened 2021-09-22
The existence tab needs to be added to the reports


njkim:
- [x] Instrument
- [x] Physical Thing
- [x] Digital Resource
- [x] Set
- [x] Group

PR #657, opened 2021-09-23
The existence scene in the reports, #651 

njkim:
The existence tab is added to the group and collection/set, #651


afs/media/js/views/components/reports/scenes/existence.js
```diff
@@ -33,142 +33,149 @@ define(['underscore', 'knockout', 'arches', 'utils/report','bindings/datatable']
...
-                    eventObservables.name = ko.observable(self.getNodeValue(eventData, `${existenceEventConfig}_influence`));
-
-                    return eventObservables;
-                }
+                })
             }
+
             // if params.compiled is set and true, the user has compiled their own data.  Use as is.
             if(params?.compiled){
                 self.eventData = params.data.eventData;
```
aarongundel, line 175: This should be updated to eventDataArray to match.

afs/media/js/views/components/reports/scenes/existence.js
```diff
@@ -33,142 +33,149 @@ define(['underscore', 'knockout', 'arches', 'utils/report','bindings/datatable']
             self.delete = params.deleteTile || self.deleteTile;
             self.add = params.addTile || self.addNewTile;
             self.events = params.events || ['production']
-            self.eventData = {};
+            //self.eventData = {};
```
aarongundel, line 36: This should be removed.

afs/templates/views/components/reports/scenes/existence.htm
```diff
@@ -6,72 +6,75 @@
...
+    <!-- ko if: eventDataArray[eventName].length === 0 || !eventDataArray[eventName].some(x => x !== undefined) -->
     <div class="afs-nodata-note">No events</div>
     <!-- /ko -->
+
     <!-- Show this div if there are event data -->
     <!-- Event Block (one for each event) -->
-    <!-- ko if: eventData[eventName] && eventData[eventName].tileid --> 
+    <!-- ko if: eventDataArray[eventName].length > 0 && eventDataArray[eventName].some(x => x !== undefined) -->
+    
+    <!-- ko foreach: { data: eventDataArray[eventName], as: "eventData" } -->
```
aarongundel, line 25: It's worth noting that setting noChildContext to true here is probably the best solution since you're already using an "as" reference - then you wouldn't need to go through and modify the parent context for all of these child elements.  Note above that where you're adding it - you shouldn't need to add it, but it works because the $parent is whatever's calling existence, which will also happen to inherit from the report.js helper.  The solution works, but I've found nested contexts painful to deal with.  Your choice on whether to change or not.  Let me know if you'd prefer keep as is and I'll resolve this comment.

aarongundel (changes requested):
Looks good - few minor issues.  Were you able to test this on physical thing/instrument also?  If it works there too, it's probably good across the board.

afs/media/js/views/components/reports/textual-work.js
```diff
@@ -34,22 +35,52 @@ define([
...
+            self.existenceCards = {};
+
+            console.log(self.resource())
 
             if(params.report.cards){
                 const cards = params.report.cards;
                 
                 self.cards = self.createCardDictionary(cards)
-
+                console.log(self.cards)
```
aarongundel, line 51: This console.log and the one on L45 should be removed.

aarongundel (approved):

Issue #656, opened 2021-09-22
Replace placeholder concepts for sample, sample area, and analysis area in workflows


chiatt:
The concepts for sample, sample area, and analysis area are now available and should replace the placeholder concepts used to define object type in the sample location and analysis areas workflows.

PR #664, opened 2021-09-23
Updates physical thing types in workflows

chiatt:
Replaces the placeholder concept valueids that are used to identify physical things as samples, sample areas, and analysis areas. Re #656


njkim (approved):
works as described

Issue #647, opened 2021-09-20
Repeating upload dataset for a parent thing creates duplicate relationships to the region


chiatt:
To reproduce:

Go through the upload dataset and create a dataset for a part of a physical thing.
Repeat the upload dataset workflow for the same physical thing. Upload additional files to the dataset created previously.
Open the dataset's physical thing (child of the parent physical thing) and observe that there are two tiles representing the relationship to the dataset.

PR #680, opened 2021-09-25
Prevent creation of extra relationship tile to dataset, re #647

chiatt:
Prevents an extra dataset from getting created when the select files for dataset step loads. re #647


njkim (approved):
Resolve the issue #647

Issue #676, opened 2021-09-24
Update motivation statement type in sample taking workflow


chiatt:
Currently when we create a sample in the sample taking workflow the motivation sample type is set to `object/work type (category)`. The `motivation` type is now available and should replace `object/work type (category)`.

PR #682, opened 2021-09-25
Replace placeholder valueid with 'sampling motivation' valueid

chiatt:
Replace placeholder valueid with 'sampling motivation' valueid. re #676


njkim (approved):
works as expected

PR #684, opened 2021-09-29
Report update substance timespan

njkim:
Update substance scene report template to support timespan

### DIFF
--- a/afs/media/js/views/components/reports/observation.js
+++ b/afs/media/js/views/components/reports/observation.js
@@ -36,7 +36,7 @@ define([
             };
             self.substanceDataConfig = {
                 dimension: undefined,
-                timespan: {nodegroup: 'timespan', key: 'timespan of observation'}
+                timespan: {path: 'timespan', key: 'timespan of observation'}
             };
             self.nameCards = {};
             self.descriptionCards = {};
--- a/afs/media/js/views/components/reports/observation.js
+++ b/afs/media/js/views/components/reports/observation.js
@@ -34,9 +34,14 @@ define([
                 'subjectOf': undefined,
                 'label': undefined
             };
+            self.substanceDataConfig = {
+                dimension: undefined,
+                timespan: {nodegroup: 'timespan', key: 'timespan of observation'}
+            };
             self.nameCards = {};
             self.descriptionCards = {};
             self.documentationCards = {};
+            self.substanceCards = {};
             self.summary = params.summary;
 
             if(params.report.cards){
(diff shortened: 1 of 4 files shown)


afs/media/js/views/components/reports/scenes/substance.js
```diff
@@ -27,42 +27,107 @@ define(['underscore', 'knockout', 'arches', 'utils/report','bindings/datatable']
...
-                self.dimensions(params.data.dimensions);
-            } else {
-                let dimensionData = self.getRawNodeValue(params.data(), self.dataConfig.dimension);
-                
-                if(dimensionData) {
-                    if(dimensionData.length === undefined){
-                        dimensionData = [dimensionData];
-                    } 
+            if(self.dataConfig.dimension){
+                if(params?.compiled){
```
aarongundel, line 31: Seems like params.compiled should be outside of this if check - if params.compiled is set, the user has compile/provide to set _all_ of the data on their own, including dimension and timespan.

afs/media/js/views/components/reports/observation.js
```diff
@@ -34,9 +34,14 @@ define([
                 'subjectOf': undefined,
                 'label': undefined
             };
+            self.substanceDataConfig = {
+                dimension: undefined,
+                timespan: {nodegroup: 'timespan', key: 'timespan of observation'}
```
aarongundel, line 39: I'd rename 'nodegroup' to 'path' since that's how we're referring to this in scenes.

afs/media/js/views/components/reports/scenes/substance.js
```diff
@@ -27,42 +27,107 @@ define(['underscore', 'knockout', 'arches', 'utils/report','bindings/datatable']
...
+                    }
                 }
+            }
 
-            } 
+            self.timeSpanValue = ko.observableArray();
+            self.timeSpanData = ko.observableArray();
 
+            if(self.dataConfig.timespan){
+                if(params?.compiled){
```
aarongundel, line 71: same comment re: params.compiled

aarongundel (changes requested):
Looks good!  Couple of minor comments

afs/media/js/views/components/reports/observation.js
njkim: Renamed

afs/media/js/views/components/reports/scenes/substance.js
njkim: Removed

afs/media/js/views/components/reports/scenes/substance.js
njkim: Rearranged, thanks

aarongundel (approved):
looks good

PR #687, opened 2021-09-30
new label based graph based reports

aarongundel:
Adds new label based graph based reports


chiatt (changes requested):
This looks really good - there are just a couple of issues that I ran into. The first is that the substance scene breaks when there is no data:
image

Another issue is that in the report page - outside of the editor, the parts scene breaks with the following error: 
image

aarongundel:
thanks for that!  Fixed.

chiatt (commented):
Looks really good

Issue #674, opened 2021-09-24
Add name_language to Instrument name 


chiatt:
Add the default name_language to Instrument name in the second step of the upload dataset workflow

chiatt:
Fixed in #689

PR #689, opened 2021-10-01
Set default language on obs name and statement, re #674

chiatt:
Sets default language on observation name and the observation parameter statement, re #674


njkim (approved):
works as described

Issue #681, opened 2021-09-25
Analysis Area workflow hangs when completing workflow


chiatt:
In the second to the last step in the Analysis Area workflow, if the user clicks `Save and Complete Workflow` the page hangs with the loading spinner.

PR #691, opened 2021-10-01
Update the value with the savedData in analysis area annotation step, #681

njkim:
Fix the issue of the step's hanging, when _`saved and completed workflow`_ at the analysis area annotation step, #681 
by updating the self.value with savedData.


chiatt (approved):
Works as expected

PR #692, opened 2021-10-01
Save sample type & lang w/ name tile, #667

njkim:
Save the name with language and type, #667 
by using tile api and saved with English & [preferred term, primary title]


chiatt (approved):
Looks good to me

Issue #677, opened 2021-09-24
Workflow output name changes in sample taking


chiatt:
Currently it is 'Sample' but should be changed to 'sampling' and 'Region' prefix for the sample area name should be changed to 'Sample Area'

chiatt:
Fixed in #693

PR #693, opened 2021-10-01
Update the resource name prefixes, #677

njkim:
Update the sampling activity and sample area physical thing names properly, #677
sampling activity will be "Sampling for ..."
sample area will be "Sample Area for ..."


chiatt (approved):
Looks good

Issue #694, opened 2021-10-01
Sample taking - sample info step hangs when reivisted and saved.


njkim:
If a user revisits the sample info step (2nd step) of sample taking workflow and save again, the step hangs.

PR #696, opened 2021-10-01
Modify sampling info saving, #694

njkim:
Flip the complete flag before saving to resolve the issue with saving is not done and hangs, #694


njkim:
the issue resolved with #702

Issue #669, opened 2021-09-24
Create relationship tile between samples and their sample area


chiatt:
Using 'Removal from Object_Removed From' on the physical thing model

PR #698, opened 2021-10-04
Save the sample-sample area relation, #669

njkim:
Add a related resource tile in the sample resource instance which is referenced sample area, #669 
using Removal from Object nodegroup (Part Removal Event of Object card)


chiatt (approved):
Works as expected

Issue #672, opened 2021-09-24
Require project selection in first step of the analysis areas workflow


chiatt:
Much like the second step of the upload dataset workflow and the first step of sample taking

PR #700, opened 2021-10-05
Select project in first step of analysis areas, re #672

chiatt:
Replaces first step of analysis areas workflow with the select-phys-thing-step which requires project selection. re #672


njkim (approved):
works as expected

Issue #668, opened 2021-09-24
Add the preferred manifest to the sampling activity


chiatt:
We need to indicate on the sampling activity which physical manifest was used. This can likely be saved to the activity in step 3 of sample taking.

PR #702, opened 2021-10-05
Add the preferred manifest to sampling activity, #668 

njkim:
Add the preferred manifest chosen to the sampling activity as well as physical thing (currently just phys thing), #668


chiatt (approved):
Works as expected

PR #707, opened 2021-10-07
Afs shared containers

ryan86:


aarongundel (approved):
pretty straightforward - looks good!

Issue #646, opened 2021-09-20
Duplicate sampling unit tile created in sample location step of sample taking workflow


chiatt:
The sample physical thing has an empty sampling unit (possibly created when Overall Object Sampled tile is initially added). See example here: https://recordit.co/Pv5zguMrEF

chiatt:
The issue here seems to be that a Sampling Unit tile is created when the sampling activity instance is created. This is because part of the Sampling Unit is Overall Object Sampled. It isn't really a duplicate. It simply an extra Sampling Unit that doesn't have a sample associated with it. It does't seem necessary to create this tile here because the Overall Object Sampled is created in other Sampling Units as each sample is created.

PR #709, opened 2021-10-08
Remove sampledObject creation w/o annotation, #646

njkim:
Remove the sampledObject (which relates Physical Things to sampling) tile creation in sampling info step, #646 
Because the tile should include sample location annotation (as in the sample location step), saving the tile in sampling info step without an annotation is unnecessary.


chiatt (approved):
Works as expected

Issue #662, opened 2021-09-23
Search results of add things to set should return all results that match term 


chiatt:
If a user enters a term in search in the add things to set step, they can only select an item in the search suggestions on which to search. This results in only being able to select one item a time from the results. The term search should load all results with that term in the results so that users can select more than one item at a time.

PR #711, opened 2021-10-08
Add search term to select2 drop down selection, #662

njkim:
Add the search term itself in the select2 dropdown selection in the add-things-step in the new project workflow, #662


chiatt (approved):
Works like a charm

Issue #608, opened 2021-09-08
Create widget to manage calculated strings


chiatt:
In the analysis area and sample location workflow steps and we calculate the name of the resulting physical thing. The issue is that the value of the string widget (non-calculated) matches the value of the list of parts. So in the UI the user sees the calculated name of the physical thing, but not the calculated name in the list of parts. To correct this, we need a widget that stores a calculated value based a parameter - in this case the name of the parent physical thing.

aarongundel:
https://github.com/aarongundel/arches-for-science-prj/tree/608_tile_name_changes -> this branch has the changes we discussed re: the physical thing name list appending @chiatt

chiatt:
This may be difficult and possibly unnecessary. It may be easier and better to implement the changes that @aarongundel has in the branch above. The issue with those changes, however, is that they would create long names that are difficult to display in the workflow step.

PR #714, opened 2021-10-11
Rename analysis area, sampling activity related resource names, #608

njkim:
Rename analysis area, sampling activity, sample, sample area names. #608 
ie. if creating a sample of an area `Right Foot` from `Statue` physical thing.

Right Foot [Region of Statue]
Right Foot [Sample of Statue]
Right Foot [Sample Area of Statue]


chiatt (changes requested):
Hey @njkim, could you add the `afs-workflow-table-contents` class to the analysis areas list as well. The ellipsis is only showing in the sample location step.

Issue #743, opened 2021-10-29
Dataset upload step sometimes fails to list files in table.


chiatt:
Dataset upload step sometimes fails to list files in table. Objects appear in the annotation viewer, but are not listed in the table.

PR #744, opened 2021-10-29
Fix issue not showing the lists of files in upload data final tables, #743 

njkim:
Also, update to show only the annotation that has uploaded files. #743 
Fix with analysis area final step is included too.


chiatt (approved):
Looks really good.

PR #747, opened 2021-10-31
Support multi-canvas editing in sample taking workflow

chiatt:
Allow user to switch between canvases to edit sample locations in each. Also sets all overlays to visible by default in sample taking and analysis areas workflows and makes the color of the layer which the user is not editing to be consistent with the color for that layer in the final step.


njkim (approved):
Works great with multiple canvases.

Issue #753, opened 2021-11-01
Analysis Area Workflow FInal Step: Issue with number of areas in table


dwuthrich:
The "Regions of Interest" table shows an additional bland (and un-needed) row, which forces the counter widget too far below the table:

![Screen Shot 2021-11-01 at 1.06.22 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/4492c571-25b5-427c-ad00-182f64dc6111)

PR #759, opened 2021-11-02
Move margins out of the component, #753

njkim:
Fix the issue with showing extra gap(margin) and also not showing the count properly. #753


chiatt (approved):
Looks good - removes extra space beneath table.

Issue #725, opened 2021-10-19
New Project workflow collection step: search filter not used when paging through results


dwuthrich:
If a user filters the list of physical things to add to a project, the filter is ignored when the results are pages as shown in this screen capture: https://recordit.co/WYPoAWQyqD

PR #768, opened 2021-11-02
Fix add things step to return next page result properly, #725

njkim:
Fix paging issue in add-things-step by adding term-filter to the subsequent paging-query, #725


chiatt (approved):
Nice that you got that fixed with so few changes!

Issue #757, opened 2021-11-01
Create a default name for the uploaded dataset in the Upload Dataset workflow


dwuthrich:
Users are required to provide a name for the dataset associated with a sampling or analysis area.  This currently is just what the user enters, which will likely be something like "dataset 1".

We should post append the name of the physical thing the dataset is associated with (like we do with other names in the AfS workflows).  

What should happen:  for physical thing named "A1-A [Region of Arcadian Shepards]", the sample name for the digital resource we create in the workflow should be: Dataset 1 (A1-A [Region of Arcadian Shepards])

PR #769, opened 2021-11-03
Auto-generate the dataset name with physical thing suffix, #757

njkim:
with the format of 
`Dataset Name (Physical thing Name)`

If a user typed "XRF dataset"
e.g. XRF dataset (Right hand [Region of Statue A])

If a user decide not to add dataset name it will simply call dataset.
e.g. Dataset (Right hand [Region of Statue A])

### DIFF
--- a/afs/media/css/project.css
+++ b/afs/media/css/project.css
@@ -694,6 +694,15 @@
     text-overflow: ellipsis;
 }
 
+.afs-calculated-datasetname {
+    margin: 5px 0px 15px 0px;
+    padding-left: 15px;
+    height: 36px; 
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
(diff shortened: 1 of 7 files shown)


chiatt (commented):
This looks good - but I noticed a couple of issues. I wasn't able to save the dataset with the default name. I had to first modify the name. Also, after the name was changed, wouldn't get the dirty state if I wanted to change it later to something else.

njkim:
> This looks good - but I noticed a couple of issues. I wasn't able to save the dataset with the default name. I had to first modify the name. Also, after the name was changed, wouldn't get the dirty state if I wanted to change it later to something else.

I pushed a new fixed to the pr. Now the dirty state should be all properly activated. Thanks.

chiatt:
Works like a charm

Issue #739, opened 2021-10-29
Add Type to the digital resource created in image step


chiatt:
Type should be 'Manifest' and in the 'Digital Service' branch, the type should be 'web service'

PR #778, opened 2021-11-04
Create digital resource type / service type, #739

njkim:
The digital resource type and digital resource service type are added when the image service step creates digital resource  in analysis area & sampling workflows. #739


chiatt (approved):
Works as expected

Issue #765, opened 2021-11-02
In sample annotation step, replace spinner with progress info


chiatt:
In sample annotation step, replace spinner with progress info about what is getting saved

PR #782, opened 2021-11-05
Add saving progress info message, #765

njkim:
Analysis Area Annotation step & Sample Taking Sample Location step.
Showing the message `"Saving something ..."`


chiatt (approved):
This looks really good and I think is a much better user experience

Issue #785, opened 2021-11-05
Scrollbars appear below report tabs on Windows


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/0b508855-130f-4b42-841f-bf434c1c7d48)

chiatt:
Fixed in #786

PR #786, opened 2021-11-05
Set tab container x-axis scroll to auto for windows, re #785

chiatt:
Removes unnecessary horizontal scrollbars beneath report tabs when viewed on Windows Chrome

### DIFF
--- a/afs/media/css/report_templates.css
+++ b/afs/media/css/report_templates.css
@@ -74,7 +74,7 @@
 }
 
 .afs-report-anchor-container {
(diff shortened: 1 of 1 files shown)


njkim (changes requested):
The unnecessary scroll bar was fixed. However, even when the windows size is actually small and the scroll is desirable (or necessary), the scroll does not seem to be working...

njkim (approved):
It works.
(Somehow it did not look like it was working previously, maybe cache.)

Issue #784, opened 2021-11-05
Remove 'Switch Image Service' button in workflows


chiatt:
We can't allow a user to switch their image service during workflows, so we need to remove that option in all workflows with the annotation editor.

![Screen Shot 2021-11-05 at 6.16.11 AM.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/726162aa-0b66-4e0d-af6f-48e1c93909dc)

PR #788, opened 2021-11-06
Hide switch manifest button during workflow, #784

njkim:


njkim:
> This looks good, but it needs to be applied to the 'upload dataset' workflow as well.

Upload-dataset step is added to css to hide 'switch' button.

chiatt (approved):

Issue #758, opened 2021-11-01
Add activity type to the project info step


chiatt:
The user needs to be able to define the activity type for their project in the new project workflow. This should be added to the project info step.

PR #789, opened 2021-11-06
Add activity type to the new project, #758

njkim:


chiatt (changes requested):
It may be possible to do this by adding a workflowstepclass rather than modifying the graph.

njkim:
> It may be possible to do this by adding a workflowstepclass rather than modifying the graph.

Now utilizing the `workflowstepclass` and graph in package is back unmodified.

chiatt (approved):

PR #792, opened 2021-11-09
Allow for a single feature to be added with iiif annotations

aarongundel:
Modifies the analysis-areas-annotation-step component and the sample-taking-sample-location-step component to only allow for a single feature to be added.  Checks widget config in cards_x_nodes_x_widgets.  Updates that table and default widget template table on package load to turn of multi-feature support.  Multifeature is enabled by default.  

There will need to be an additional feature added to core arches to support this in the base component, if desired.


chiatt (approved):
Looks good to me

Issue #767, opened 2021-11-02
Summary Report does not show latest data of instance


dwuthrich:
A summary report does not necessarily show the data for its instance.  In the following example, the summary report for a group does not display the data associated with the Classification group

See: https://recordit.co/DG2DvvWq91

chiatt:
Fixed in #796

PR #796, opened 2021-11-09
Fix summary report to show exact match, #767

njkim:
Fix the issue that the summary report does not show the value properly, #767


chiatt (approved):
Fixes #767

Issue #776, opened 2021-11-03
AfS Editor Report Template: Group template has error in communication tab


dwuthrich:
The template shows the "contact points" data at the very bottom of the page:
![Screen Shot 2021-11-03 at 3 21 26 PM](https://user-images.githubusercontent.com/815185/140202299-4fbe453b-3115-4a70-a5ff-a95b2a602b13.png)

PR #798, opened 2021-11-09
Group custom report communication tab, #776

njkim:
Fix the issue showing a big gap above the contents, #776


chiatt (approved):
resolves #776

Issue #791, opened 2021-11-09
Unable to delete a geometry in sample location step


chiatt:
In the sample taking workflow deleting a geometry fails with the following error: 
`tile.data[physicalThingPartAnnotationNodeId].features is not a function`

PR #799, opened 2021-11-09
fixes issue where you cannot delete a feature

aarongundel:
Fixes issue where new (unsaved) features cannot be deleted.  After this fix, you can remove features from the feature list if they have not been saved.


chiatt (approved):
Fixes #791

Issue #777, opened 2021-11-03
AfS Editor Report Template: Parthood template only shows a single manifest


dwuthrich:
The template all the parts on the first manifest, instead of showing each manifest and its parts in separate divs, as the summary page in the analysis areas workflow does.

To reproduce:
1. Add a part to a Physical Thing using a manifest
2. Add a second part using a different manifest
3. View the result in the parthood tab and note that all the parts appear on a single image.

What should happen:
Each unique manifest/canvas combination should have its own section (e.g.: image viewer/table pair) in the parthood tab

PR #803, opened 2021-11-10
Update the custom report parthood to show multiple canvas, #777

njkim:
To show the `parts` annotation with the multiple canvases, #777 
Fix the issue of adding all the annotation geometry on one canvas


chiatt (approved):
Looks good to me!

Issue #804, opened 2021-11-11
Datasets created without files


chiatt:
In the dataset workflow, if there are multiple annotations, but only one of them has files associated with it, digital resource instances should only be created for the one with files. However, because of auto-naming, digital resources are created for all of the annotations.

PR #805, opened 2021-11-11
Prevent dataset save for parts with no files or name, re #804

chiatt:
In the upload dataset workflow, prevents the creation of a dataset for parts that do not have files or a name. re #804


njkim (approved):
fixed the issue, works great

Issue #781, opened 2021-11-04
Ensure appropriate datasets are selectable in review dataset workflow


chiatt:
Prevent selection of manifests as datasets in the review dataset workflow
Be sure newly created datasets are selectable

chiatt:
Resolved in #807

PR #807, opened 2021-11-11
Update review dataset workflow, #781

njkim:
Fix a few issues with review dataset workflow, #781
- prevent from choosing the manifest
- let user to choose the parent physical thing to start with
- show all the annotations with the datasets in the final step
...


chiatt (changes requested):
Hey @njkim, I'm still able to select my manifest in the first step, but I think that should be prevented in this PR.

chiatt (approved):
Works well for me

Issue #802, opened 2021-11-10
Annotation view width is hardcoded in summary steps


chiatt:
The annotation viewer width is hardcoded at 400px causing it to fill only a fraction of its parent element in the summary page. This appears to be a problem in the summary steps for the sample taking, analysis areas, and upload dataset workflows.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/ab8fe77d-7840-4057-9503-b2fe27ac7e67)

PR #809, opened 2021-11-12
Update the annotation width Upload dataset final step, #802

njkim:
update the annotation width (leaftlet container width) to 100%, #802


chiatt (approved):
Looks good

Issue #808, opened 2021-11-12
Minor UI edits


chiatt:
- New project workflow, first tab: Change the helper ‘step information’ text to be “Identify the project by giving it a name” – we don’t have the objectives on this tab, so want to remove that instruction.
- New project workflow, second tab: Change the helper ‘step information’ text to be “Indicate the objectives or motivation of the project” rather than the generic ‘set the project statement’ (Aside – is this really required? I didn’t realize that….)
- New project workflow, third tab: Change the helper ‘step information’ text to be “Indicate the date the project started if known”
- New project workflow, third tab: Change the helper ‘step information’ text to be “Project Team: Add the names of people involved in the project” (currently it’s about timespan….)
- Sample taking WF, first tab: grammar issue in ‘step info’ text – should be “you are sampling” instead of “your are sampling”
- Upload dataset WF, first tab: for first radio button, change “on remote samples” to “on removed samples”.
- Upload dataset WF, fourth tab: The tab itself should be called something like ‘data files and locations’. It could be slightly confusing to start with uploading files when the tab indicates location, which is actually the second thing happening on the tab….
- Sample taking WF, second tab: can first box be motivation, second one be technique?

chiatt:
Resolved in #812

PR #812, opened 2021-11-12
Minor workflow UI edits

chiatt:
Minor UI edits to workflows #808


njkim (approved):
The descriptions are more informative and make more sense. Good.

Issue #813, opened 2021-11-12
In the upload dataset workflow, a user should be able to choose a sample


njkim:
The issue:
If a user chooses the 2nd option (Data from the consumption of the sample or object) then chooses one of the samples, "The selected physical thing is invalid" message appears.

The desired behavior:
Since the 2nd option assume the consumption of the sample, the user should be able to choose the sample. But probably only the sample not the analysis area or sample area.

chiatt:
resolved in #820

PR #820, opened 2021-11-13
Include sample to select when destructive, #813

njkim:
The samples w/o annotation can be selected for the destructive observation, #813 
Fix the issue of not being able to choose sample (if the sample does not have annotation) when physical thing is selected in the upload dataset workflow.


chiatt (approved):
This works, but it seems like we still need to filter out all of the sample areas and analysis areas from the dropdown, so that users only have samples to pick from. Either that or we should update the error message so that the user understands why selecting an analysis area failed. However, only offering samples for selection would be a better user experience.

PR #822, opened 2021-12-03
adds delete without redirect for reporting

### LINKED ISSUE #772: AfS Editor Report Template: user should stay on report when deleting a tile
When deleting a tile from instance report, the page refreshes itself with the branch (from which the tile was deleted) now active.

What should happen: the page should refresh but stay on the same report tab that the user was on.

https://recordit.co/mmTRBlRUdF

aarongundel:
resolves #772 

Should be migrated into aher after merge.

### DIFF
--- a/afs/media/js/bindings/datatable.js
+++ b/afs/media/js/bindings/datatable.js
@@ -57,16 +57,7 @@ define([
                 table.destroy();
             }
             ko.bindingHandlers.foreach.update(element, valueAccessor, allBindings, viewModel, bindingContext);
-            const opt = options.dataTableOptions;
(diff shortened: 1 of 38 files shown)


aarongundel:
found issue with one of the graphs; don't merge until all clear.

aarongundel:
fixed issue with deleted rows reappearing.  Maintains list of deleted rows in the report viewmodel.  That's checked by the datatable binding.  If it's found (the tileid), the row isn't shown.

chiatt (approved):
Looks good. There does seem to be an issue with the UI updating on delete of a single cardinality n tile, but that issue is likely due to a bug in the label based graph and is addressed by #828

PR #823, opened 2021-12-03
fixes issue with no data block always appearing

### LINKED ISSUE #775: AfS Editor Report Template: Existence template shows "no data" block incorrectly
In the Physical Thing model, users can add parts to a production event.  In the existence tab, these data are shown in the Event parts section of the template.

Currently, a message saying "no time span data for this event" is shown, even when the data exist
Screen Shot 2021-11-03 at 3 17 55 PM
:

aarongundel:
resolves #775 

The "no data" block was not wrapped in anything to prevent it from appearing.


chiatt (approved):
Looks good

PR #824, opened 2021-12-06
inline the key/value pairs in some scenes

### LINKED ISSUE #773: AfS Editor Report Template: move related resource position
Currently, the "--" used to indicate no value and the resource name for related resource templates should appear betwen the label and "add/edit" link as shown here: https://recordit.co/3oVZazVENs

aarongundel:
Fixes #773 

Inlines various key/value pairs as requested by Dennis.  

Also fixes a couple of bugs in the existence scene that I uncovered while fixing the key/value pairs.


chiatt (approved):
Looks good to me

Issue #829, opened 2021-12-20
Error when viewing parthood scene in report


chiatt:
Parthood loads as expected in the editor, but not a standard report

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/eea63d8d-659c-44f5-9d01-bb3d634549e7)

chiatt:
Fixed in #830

PR #830, opened 2021-12-21
Fix the issue with annotation custom report, #829

njkim:
Fix the issue with annotation custom report, #829
By making the annotation info an observable


chiatt (approved):
Looks great

Issue #831, opened 2021-12-21
Hide visual work node in Phys Thing parts branch


chiatt:
Hide visual work node in Phys Thing parts branch

chiatt:
Fixed in #832

Issue #800, opened 2021-11-09
Remove default manifest from physical thing parts annotation node


chiatt:

chiatt:
This has been applied to the demo branch, but still needs to be applied to master following the upcoming demo.

chiatt:
Resolved in #800

PR #832, opened 2021-12-21
Remove default manifest from phys thing part and hide visual work widget, #800, #831

chiatt:
Removes  the default manifest from physical thing part annotation node and hides the visual work widget, #800, #831


njkim (approved):
Works as described

Issue #839, opened 2021-12-22
Show only display name in search suggestions (Hide model and node name)


chiatt:
Search suggestions show the display name suffixed by the model name and the node associated with the term. In the case of the 'Add objects to project' step of the new project workflow, this is not helpful because all suggested terms belong to the same resource model. It would be helpful to hide this information and show only the display name. 

This fix is probably something we need to do in core arches

PR #875, opened 2022-01-11
Modify create-project-workflow add-things-step search suggestions, #839

njkim:
Show only display text for search suggestions, #839


chiatt (approved):
Nice improvement. Glad changing core was not needed.

PR #876, opened 2022-01-11
preexisting sample fixes

### LINKED ISSUE #858: Prevent users from modifying pre-existing samples in sample taking workflow
Currently the annotation step of the sample taking workflow allows a user to make changes to pre-existing samples. This is problematic because the edited samples are not associated with the sampling activity instance created in the current workflow. Users should not be allowed to make changes to pre-existing samples, but instead should only be allowed to change the samples created in the current instance of the workflow.

re archesproject/arches-for-science-prj#730

aarongundel:
Disables preexisting samples locations tied to another sampling activity.  

Resolves #858


aarongundel:
Needs https://github.com/archesproject/arches/pull/8078

aarongundel:
Per feedback from @chiatt  - added a lock icon to rows of samples that are associated with other sampling activities.  Also recreated the check for an invalid samplingActivitySamplingUnitCard.  Removed console logging.

chiatt (approved):
Looks good to me

PR #878, opened 2022-01-13
fixes issue with tying datasets to pre-existing observations

### LINKED ISSUE #859: Prevent users from modifying datasets that do not belong to the current observation
Currently in the upload datasets step in the upload dataset workflow, users can edit datasets associated with other observations. This is problematic because changes might then not be associated with the correct observation but rather just the current observation. Users should only be able to modify datasets created during the current workflow.

aarongundel:
You can no longer tie datasets to multiple observations

fixes #859


chiatt (approved):
Looks great

PR #879, opened 2022-01-14
remove rows from existence scene when they're null

### LINKED ISSUE #774: AfS Editor Report Template: Existence template to show/hide information correctly
The existence template currently shows the production event names/statements/identifiers/timespans even when no data exists for these cards

To reproduce:
1. Create new Physical Thing
2. Create a new Production Event tile by adding information to the top card of the branch (but no child data)
3. View the Existence tab in the report and note the presence of data blocks with no information

Example
Screen Shot 2021-11-03 at 3 12 37 PM
:

aarongundel:
removes rows from the existence scene when those rows are "null" - have no displayable values.

Fixes #774


njkim (approved):
It looks good.

PR #880, opened 2022-01-17
crud operations in upload workflow

aarongundel:
Modifies the upload workflow to delegate save/undo/cancel operations to the workflow buttons (previously hidden).  I did not remove the save/cancel buttons - as I could see a use case where a user might want to save/cancel and not advance the workflow.  But that should be easy enough to do if desired.  This is done for the select files step and the file interpretation step, both of which needed the change.  Also fixes state issues (savedData/value) that were causing this not to work beforehand.


aarongundel:
Assigned you for review @njkim - if that's not appropriate feel free to reassign.

chiatt:
Closing because interface will be refactored

Issue #834, opened 2021-12-22
Create workflow to add additional objects to a project


chiatt:
This could be a basic subset of steps in the new project workflow

PR #881, opened 2022-01-19
A new workflow to update the project collection, #834

njkim:
Add a new workflow to update physical things in the project collection, #834


afs/plugins/init-workflow.json
```diff
@@ -15,6 +15,15 @@
                 "circleColor": "#b0adf3",
                 "desc": "A research effort, study, or conservation project"
             },
+            {
+                "workflowid": "ef3a374c-f795-4317-aaee-83e1493eabb6",
+                "slug": "project-collection-workflow",
+                "name": "Update Project Collection",
```
chiatt, line 21: Could the name be changed to "Manage Objects in a Project"?

afs/plugins/init-workflow.json
```diff
@@ -15,6 +15,15 @@
...
+                "icon": "fa fa-files-o",
+                "bgColor": "#3e89cd",
+                "circleColor": "#6ea7d9",
+                "desc": "Review digital data files generated by a lab instrument"
```
chiatt, line 25: Update description to read "Add objects to an existing project"

chiatt (changes requested):
@njkim This looks pretty good. I made a couple of inline comments. Also, I noticed that the object selection step does not display the way that it does in the Create Project workflow. The paginator is not positioned at the bottom of the step.

image

afs/plugins/init-workflow.json
njkim: updated

afs/plugins/init-workflow.json
njkim: updated

njkim:
> @njkim This looks pretty good. I made a couple of inline comments. Also, I noticed that the object selection step does not display the way that it does in the Create Project workflow. The paginator is not positioned at the bottom of the step.

Fixed

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
```diff
@@ -102,6 +106,37 @@ define([
...
+
+            const collectionRelatedResources = await (await window.fetch(`${arches.urls.related_resources}${self.collectionResourceId()}`)).json();
+            const childPhysicalThingsValueIds = [
+                '77d8cf19-ce9c-4e0a-bde1-9148d870e11c', //sample
+                '7375a6fb-0bfb-4bcf-81a3-6180cdd26123', //sample location
+                '31d97bdd-f10f-4a26-958c-69cb5ab69af1', //anlysis area
+            ];
+
+            self.startValue(
+                collectionRelatedResources.related_resources.related_resources
```
chiatt, line 121: @njkim We should confirm with @dwuthrich whether sample / analysis area instances should appear in this list or not.

chiatt (approved):
Looks great. One thing we will need to follow up on is whether samples and analysis areas should appear with the parent in the collection list.

Issue #877, opened 2022-01-12
.bmp files do not display in interpretation step in 'destructive' dataset upload workflow


chiatt:
.bmp images do not display in file interpretation step when rendered in the 'destructive' path of the dataset upload workflow.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/81395927-8ec0-4829-a4a9-4f992d0c80dd)

PR #882, opened 2022-01-19
Add image renderer file upload step - upload dataset workflow, #877

njkim:
Update the renderer assignment to add `image` and `pdf`, #877


chiatt (approved):
Looks good to me

Issue #762, opened 2021-11-02
Dataset upload workflow: Improve UI when saving dataset/files 


dwuthrich:
In the Sample Location step of the Upload Dataset workflow, users can associate files to specific datasets.

When saving their work, Arches begins saving each dataset and eventually shows a green arrow next to each file that has been uploaded.  This is nice.  However, for the case when a user is creating many datasets with many files, the UI can appear to be stuck (because the accordian widget may be hidden for the dataset that is being processed and therefore hidding the placement of green arrows next to each file).

What should happen: the UI should show a message stating that the files are being uploaded

PR #883, opened 2022-01-20
Add progress message - upload dataset workflow, #762

njkim:
Add progress message in the select dataset files step in upload dataset workflow, #762

### DIFF
--- a/afs/media/css/project.css
+++ b/afs/media/css/project.css
@@ -891,11 +891,3 @@
     position: absolute;
     z-index: 100000001;
 }
-
-div.workbench-card-sidepanel .afs-saving-message {
-    font-size: 16px;
-    color: #333;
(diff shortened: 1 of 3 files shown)


chiatt (changes requested):
Just to be consistent, I think this should look like the loading spinner for the annotation step of the sample taking workflow and the annotation step of the analysis area workflow. It's probably best that the user does not try to interact with the page while files are uploading anyhow.

njkim:
> Just to be consistent, I think this should look like the loading spinner for the annotation step of the sample taking workflow and the annotation step of the analysis area workflow. It's probably best that the user does not try to interact with the page while files are uploading anyhow.

This sounds good.
Update the message as same as other saving messages.

chiatt (approved):
Looks great!

Issue #838, opened 2021-12-22
Hide samples, analysis areas in 'add things to set' search in new project workflow


chiatt:
Users should only be able to search for and add parent objects to their project's set.

PR #886, opened 2022-01-21
Update new project workflow add thing step, #838

njkim:
To limit the selection search results to show only the parent physical thing, #838 
Hide sample, sample area, region (analysis area) from the list


chiatt (approved):
Looks good!

Issue #833, opened 2021-12-22
Show only project members in dropdowns


chiatt:
Annotator dropdown in sampling and analysis areas workflows lists all persons. Instead the list should pre-populate with persons associated with the project.

PR #890, opened 2022-01-31
A new component -related resources filtered by other node values

njkim:
Add the new component extending resource-instance-widget to only show related node values, #833
In order to show the only project team members in the `person` resource instance nodes.
Sampler in sample taking workflow and annotator in analysis area workflows are also updated.


chiatt (approved):
Looks great! Works as expected in both sample taking and analysis areas workflows.

Issue #891, opened 2022-01-31
Create project workflow 1st step creates two resources 


njkim:
The type node creates its own separate resource whether or not it is filled

chiatt:
Resolved in #893

PR #893, opened 2022-02-01
Update new project name step, #891

njkim:
Update 1st step(name & type) create project workflow as a custom step, #891
In order fix the issue of creating two resources with name & type separately.


chiatt (approved):
Looks great - nice that you cleaned out the 'research activity' cruft!

PR #894, opened 2022-02-01
Add quitUrl to workflows config

njkim:
Add the quitUrls for workflows, in case the core Arches quitUrl is search home page.
archesproject/arches#8125


chiatt (approved):
Looks good

Issue #898, opened 2022-02-23
Remove references to statement_source node in workflows


chiatt:
The statement source node has been replaced with the statement creation node. Hard-coded references to the statement_source nodeid ('033578a8-1d9d-11eb-a29f-024e0d439fdb') need to be removed to prevent the sampling activity workflow from failing.

chiatt:
Resolved in #899

PR #899, opened 2022-02-23
Remove references to statement source node. re #898

chiatt:
Remove references to statement source node so that workflows continue to work after statement source node is replaced with statement creation. re #898


njkim (approved):
works as expected

PR #902, opened 2022-02-24
changes for #8177 in arches core - split panel hiding in manifest mgr

aarongundel:
changes render context whenever manifest manager appears in a workflow to hide the split panel tools.


aarongundel:
switch to #8183 in arches core to test

chiatt (approved):
Looks great

PR #904, opened 2022-03-29
Standardize workbench sidepanel according to archesproject/arches#8306

chiatt:
Updates workbench side-panel's according to archesproject/arches#8306 to keep side-panel headers positioned relative to side-panel content.


njkim:
works great

Issue #911, opened 2022-05-27
Remove unused functions 


chiatt:
The file-to-tiff and name-select-models functions are deprecated and need to be removed from AfS.

chiatt:
Completed in #978

PR #912, opened 2022-05-27
Removes unused functions, re #911

chiatt:
Removes unused functions, re #911


chiatt:
This applies to old 6.1 schema. Submitting new PR for 6.2

Issue #897, opened 2022-02-08
Manage objects in project workflow does not show all objects in th collection


dwuthrich:
The Manage Objects workflow only shows the original objects selected when running the New Project workflow.

However, a user automatically adds new objects to the collection associated with a project when they define sample or analysis areas.  These members of the collection are not shown when a user runs the manage objects workflow, as shown in the video below.

To reproduce the issue:
1. Complete the New Project workflow
2. Complete either the Sample Taking or Analysis Areas workflow
3. Run the Manage Objects workflow and select the project you created in step 1
4. Observe that the sample(s)/analysis area(s) that you created in step 2 (and which are members of the collection for the project) do not appear in the right hand panel in the manage objects step of the workflow

https://recordit.co/nXCnNF1Rqu

aarongundel:
@chiatt to confer with scientists on behavior.

PR #917, opened 2022-06-22
Show child things add things step, #897

njkim:
It is to show the added things including child things (such as area, sample, region) as well as the parent physical thing in the manage project collection workflow - add things step. #897


chiatt (approved):
Looks good and works as expected

Issue #843, opened 2021-12-23
In image steps, add thumbnail of first canvas image


chiatt:
In list of existing image services, would be nice to see a thumbnail representation of the front/main image in this list.

chiatt:
Fixed in #924

PR #924, opened 2022-07-06
Add thumbnail to sampling image step, #843

njkim:
Add thumbnails to the image service selector in the image steps, #843 
sample taking workflow
analysis area workflow

### DIFF
--- a/afs/media/css/project.css
+++ b/afs/media/css/project.css
@@ -650,6 +650,16 @@
     margin-top: -8px;
 }
 
+.afs-image-service-instructions input,
+.afs-image-service-instructions label {
+    margin: 0px 5px 0px 0px;
+    cursor: pointer;
+}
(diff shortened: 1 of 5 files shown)


chiatt (changes requested):
This works great Namjun. I did notice that the text and radio button are misaligned because of some upper and lower margins. 

image

Could you center them?

njkim:
Thanks.
Re-centered them also moved styles tags to css.

> This works great Namjun. I did notice that the text and radio button are misaligned because of some upper and lower margins.
> 
> image
> 
> Could you center them?

chiatt:
Thanks Namjun - It's mostly the text and radio button that look misaligned to me. I was thinking it should look more like this:

image

chiatt (approved):

Issue #850, opened 2021-12-27
Add date to automatically generated observation name


chiatt:
In Upload dataset workflow, a date should be added to the automatically generated observation name. This likely means that the 'timespan' card of the observation needs to be populated by either an additional step or a new input in the instrument step.

chiatt:
Fixed in #925

PR #925, opened 2022-07-07
Add date to the observation name, #850

njkim:
Add date to the observation name upload dataset workflow, #850
Also add timespan to the observation info step


chiatt (approved):
Looks good - works as expected

PR #928, opened 2022-07-14
Compose v7 fixes for AFS

aarongundel:
remove unneeded yarn installs; install requirements.txt, add settings_local when it isn't available


chiatt (approved):
Looks good!

Issue #947, opened 2022-08-31
Allow new instrument creation in the Upload Dataset workflow


aarongundel:
Currently, you cannot add a new instrument in the Upload Dataset workflow (Observation Info Step) - you must exit the workflow and/or open a new tab to enter the instrument information and then refresh the page.  It would be better if this worked like the "Procedure Used" - allowing you to enter a new procedure if there isn't one already.

chiatt:
I don't think we want to allow users to do this. This would allow the user to create an instrument that the second half of the workflow will not know how to interpret. Granted, a user could create a new instrument in the resource editor, but it should not be so easy as to be able to do it in the workflow.

PR #951, opened 2022-09-02
Allow a user to create instrument during upload dataset workflow, #947

njkim:
Allow a user to create instrument during upload dataset workflow, #947


aarongundel:
We ended up wanting to make it a little more difficult to create these instruments, so this is no longer necessary.

Issue #933, opened 2022-08-25
Project Report changes


aarongundel:
### Overall Tab Changes
- Move Description tab to first
- Change “names and classifications” to “Names, Identifiers, Classification”
- Change “substance” to “project timeline”
- Change “parthood” to “parent project”
- Remove the "Parameters/Outcomes" tab
- Remove the "Temporal Relations" tab

### Description Tab changes
- Include Statements/Notes
- Move "Type" from the old "Parameters and Outcomes" tab here
- Move "Project Team" from old "Parameters and Outcomes tab here

### Project Timeline Tab changes
- Move current contents of "Temporal relations" here.

PR #959, opened 2022-09-07
Project report changes

johnatawnclementawn:
Move and consolidate some content on project report tabs, #933.


aarongundel (changes requested):
Looks good, just a minor nit that seems incorrect in the original ticket - Also Let's modify project.js line 70 and change the title of the parthood data since we're replacing the tab named as such.  Change it to "Parent Project" - that'll be redundant, but parthood isn't very clear.

aarongundel (approved):
Looks good!

Issue #938, opened 2022-08-26
Truncate Descriptions in Create New Project workflow


aarongundel:
Currently in the new project workflow, Physical Thing descriptions are not truncated.  As a result, they can push down content on the page and end up being quite long.  These should be truncated to some arbitrary maximum of text to preserve screen real estate.

PR #962, opened 2022-09-08
Truncate descriptions in new project workflow

johnatawnclementawn:
There are two instances of physical object descriptions and one instance of a project statement that previously overflowed the div's that contained them. These css changes contain those paragraph elements. #938

### DIFF
--- a/afs/media/css/project.css
+++ b/afs/media/css/project.css
@@ -178,7 +178,7 @@
 }
 
 
-.wf-multi-tile-card-info-details p {
+.wf-multi-tile-card-info-details {
     max-height: 100px;
     max-width: 150px;
     white-space: nowrap;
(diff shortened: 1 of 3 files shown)


aarongundel (changes requested):
This won't always work as expected.  The paragraph element is generated from the rich text editor.  If you have more than one paragraph element, then your max width/height apply to all paragraph elements and the same issue will occur.  The overflow should occur at a higher level (the containing div).

johnatawnclementawn:
I tried my hardest to use percentages rather than vw, but failed to implement them because percentages refer to the parent container and the text content is buried 3+ levels underneath the divs workflow-search-list-container & wf-multi-tile-step-list, which allocate space on the page.

aarongundel (approved):
much better.

Issue #945, opened 2022-08-31
Cannot refresh sample taking workflow Sample Info step


aarongundel:
There is a problem when using the sample taking workflow and refreshing the sample info step after entering a sample.  

1.  Go through the sampling workflow to the sample info step. 
2.  Create a new Sample - you should see that the workflow will allow you to move forward.  
3. Refresh the page.  Note that the workflow does not allow you to move forward anymore.
4. Try to edit the existing sample.  Note that it will not save anymore.  It appears that the annotations for the sample are purged on refresh.
5. Try to add a new sample.  You'll find that it doesn't work.

aarongundel:
<img width="1782" alt="Screen Shot 2022-08-31 at 12 34 52 PM" src="https://user-images.githubusercontent.com/124614/187753857-a9078ec4-323e-4bb3-b3b0-719cf44e42f9.png">

chiatt:
Fixed in #965

PR #965, opened 2022-09-09
Update Sampling Activity Annotation step, #945

njkim:
Fixes issue of the annotation not loading properly on refresh also not saving after refresh, #945 
Also fix the issue with the next step button disabled.

Dependent on archesproject/arches#8935


chiatt (approved):
Looks good and works as expected.

Issue #935, opened 2022-08-26
Collection or Set Report Changes


aarongundel:
### Overall Tabs
- Change “names and classifications” to “Names, identifiers, classification”
- Move “descriptions” to second position
- Change “existence” to “timeline” 
- Add a "Related Projects" tab
- Add a "Publications" tab

### Names, Identifiers, Classification tab
- Move contents of ‘documentation’ here

### Related Projects tab
- List projects that are related resources

### Publications tab
- List all related textual work resources

PR #976, opened 2022-09-15
Update collection / set report, #935

njkim:
Update collection / set report, #935

Rearrange the tabs.
Move related project and textual reference to its own tab


aarongundel (changes requested):
@njkim this looks pretty good, I think we should change the related project tab to be a table (and rename the tab/header to related **projects.**  It seems like there can be multiple related projects, although I can't recall how that relationship works in the workflows.  LMK if you think differently.

njkim:
The related projects node are interesting one. Because the cardinality = 1 but it is still a list node (resource instance list).
So, the related projects tab appears as mixture of what is typically done for cardinality 1 and n.

aarongundel (approved):
looks good!

aarongundel (approved):
lgtm

Issue #961, opened 2022-09-08
New tab for the Project Report


aarongundel:
Add a tab called "Component Projects" and list all subprojects (a project for which the current project is the parent).  The content will be that list.  This will require a new view/api call to query for those relationships, since they will not be available on the current project resource.

PR #977, opened 2022-09-15
Add component projects tab

johnatawnclementawn:
The Fetch API is used to access all resources that are related to the current resource. Then, those resources are filtered to find only those that are directly related. Those results are then filtered even further to find only the child resources. Finally those results populate a table. #961


afs/media/js/views/components/reports/project.js
```diff
@@ -41,6 +41,40 @@ define([
...
+
+            self.collectionOfChildProjects = ko.observableArray();
+            self.visible = {
+                childProjects: ko.observable(true),
+            };
+
+            const resourceId = ko.unwrap(self.reportMetadata).resourceinstanceid;
+            const relatedResources = async() => {
+                const result = await reportUtils.getRelatedResources(resourceId);
+                const rrs = result?.related_resources;
```
aarongundel, line 57: Avoid using abbrevations - that will cause confusion.  Longer variable names are OK as long as it's clear what you're refering to.  saying const relatedResources is preferred here.  Your function name should probably be changed to something like loadRelatedResources - since the function is performing some task, and it's unclear what task that is from the name alone.

afs/templates/views/components/reports/project.htm
```diff
@@ -73,6 +73,51 @@ <h1 class="afs-report-toolbar-title"><span class="afs-report-name">{% trans "Pro
...
+                                            </tr>
+                                        </thead>
+                                        <tbody data-bind="dataTablesForEach: { data: collectionOfChildProjects, dataTableOptions: childProjectTableConfig }">
+                                            <tr>
+                                                <td data-bind="text: displayname"></td>
+                                                <td data-bind="text: displaydescription" class="td-display-description"></td>
+                                                <td class="afs-table-control">
+                                                    <div data-bind="if: link">
+                                                        <a data-bind="attr: { href: link }">
+                                                            <i class="fa fa-globe"></i>
```
aarongundel, line 107: on second thought, instead of the globe, let's use this - https://fontawesome.com/v4/icon/link

afs/media/css/project.css
```diff
@@ -864,4 +864,11 @@
     text-align: center;
     position: absolute;
     z-index: 100000001;
+}
+
+.afs-report-page .td-display-description {
```
aarongundel, line 869: avoid including HTML tag names in the CSS class name

afs/media/js/views/components/reports/project.js
```diff
@@ -41,6 +41,40 @@ define([
...
+                let childProjects =[];
+
+                const relationships = result?.resource_relationships;
+                for (indv_relationship in relationships) {
+                    if (relationships[indv_relationship]?.resourceinstancefrom_graphid ===
+                        relationships[indv_relationship]?.resourceinstanceto_graphid) {
+                            childProjects.push(relationships[indv_relationship]?.resourceinstanceidfrom);
```
aarongundel, line 64: You need not use a for at all loop here.  Running filter/map on the relationships array defined on line 60 will give you the array you need with less code.

afs/media/js/views/components/reports/project.js
```diff
@@ -41,6 +41,40 @@ define([
...
+                    };
+                }
+
+                self.collectionOfChildProjects(rrs.filter(rr =>
```
aarongundel, line 68: re: rr - avoid using abbreviations.

afs/media/js/views/components/reports/project.js
```diff
@@ -41,6 +41,40 @@ define([
...
+                    if (relationships[indv_relationship]?.resourceinstancefrom_graphid ===
+                        relationships[indv_relationship]?.resourceinstanceto_graphid) {
+                            childProjects.push(relationships[indv_relationship]?.resourceinstanceidfrom);
+                    };
+                }
+
+                self.collectionOfChildProjects(rrs.filter(rr => 
+                    childProjects.includes(rr.resourceinstanceid)));
+                
+                for (child of self.collectionOfChildProjects()) {
```
aarongundel, line 71: this is better done with array.map.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

afs/templates/views/components/reports/project.htm
```diff
@@ -73,6 +73,51 @@ <h1 class="afs-report-toolbar-title"><span class="afs-report-name">{% trans "Pro
...
+                        <!-- ko if: collectionOfChildProjects().length -->
+                        <div class="afs-report-subsection" >
+                            <div>
+                                <div class="afs-table pad-btm">
+                                    <table class="table table-striped" cellspacing="0" width="100%">
+                                        <thead>
+                                            <tr class="afs-table-header">
+                                                <th>{% trans "Project" %}</th>
+                                                <th class="min-tabletl">{% trans "Description" %}</th>
+                                                <th class="none">{% trans "Visit Resource" %}</th>
```
aarongundel, line 96: This extra column needs to be removed - your th count needs to match your td count.  Check your chrome console - there's an error in there being caused by this.

aarongundel (changes requested):
Looks good overall, few minor changes.

aarongundel (changes requested):
Minor nit with header text.  Also, remove html tags from description.  Other than that, looks good.

aarongundel:
@johnatawnclementawn can you resolve the merge conflicts here?

johnatawnclementawn:
@aarongundel resolved conflicts

aarongundel (approved):
looks good!

Issue #954, opened 2022-09-02
Port complex save logic in instrument-info-step to django view


chiatt:
Currently the logic used to save a dataset in this step is complex requiring multiple trips to the server before save completes. Moving much of this logic to a custom Django view would increase the speed and stability of this request.

chiatt:
resolved in #980

PR #980, opened 2022-09-16
centralize save logic

apeters:
re #954


apeters:
@chiatt when form errors, I'm not sure how to remove the loading mask and get back to the form

chiatt (approved):
Looks good and works as expected.

Issue #972, opened 2022-09-14
Port complex save logic in Regions step to django view


chiatt:
Currently the save logic in the regions step of the analysis areas workflow is complex requiring a series of chained ajax requests. This logic should instead be moved to a django view and handled with perhaps only one request.

chiatt:
Resolved in #983

PR #983, opened 2022-09-21
Analysis area annotation saving django view, #972

njkim:
Port the annotation steps in analysis area and sampling activity workflows.

This pr is dependent on the core arches change in
https://github.com/archesproject/arches/pull/9003


apeters (changes requested):
It looks like the following data isn't being saved (compared to what was being saved previously).  Maybe these are default values that aren't getting picked up?

Physical Thing (Sample)
- Name_Type
- Name_Language
- Statement_language 

Physical Thing (Sample Area)
- Sample Name_Type
- Sample Name_Language

njkim:
@apeters Thanks.
The name type, language and the statement language defaults are added to the tiles.

apeters (changes requested):
After saving an annotation, the next step of the workflow is not available and the user can't move to the next step.
Screen Shot 2022-10-03 at 4 11 22 PM

apeters:
After testing a bit more, things seem to work as desired.

apeters (approved):
this works as designed

Issue #994, opened 2022-09-26
Sample Taking, Analysis Area Workflows: Widget listing sample taking locations should be interactive


dwuthrich:
The widget listing sample taking or analysis area locations should zoom the image to the related geometry when the user clicks on a row in the widget.  Currently, a user cannot interact with a row, as shown in this screen capture: https://recordit.co/EFfj1hdl5H

What should happen:
row in widget highlights when user's cursor hovers and cursor becomes a pointer.  Clicking the row should zoom the canvas to the annotation geometry.

PR #999, opened 2022-09-28
Update annotation steps in sampling, analysis area workflows, #994

njkim:
Update annotation steps in sampling, analysis area workflows, #994 
1. Now a user can click the list to see the annotation in the image
2. Update the ui for previous samples (ones that are not permitted to modify)


apeters (approved):
works as expected

Issue #936, opened 2022-08-26
Observation Report changes


aarongundel:
### Overall tab changes
- Change “names and classifications” to “Names, identifiers, classification”
- Move “descriptions” to second position
- Add a "related objects" tab
- Add a "data" tab
- Change “substance” to “dates”
- Remove “location” tab
- Change “parthood” to “parent project”

### Descriptions tab changes
- Move contents of ”parameters and outcomes” to this tab EXCEPT “recorded value of observation” - see below

### Related Objects tab
- Table with all physical things related resources
-  ~Include the ‘spot map’ here – annotated image w/ with the regions of interest noted like it appears on the summary page of the ‘analysis areas’ workflow~ - moved to https://github.com/archesproject/arches-for-science-prj/issues/1000

### Data tab
- “recorded value of observation” goes here

johnatawnclementawn:
Claimed

johnatawnclementawn:
Claimed

PR #1001, opened 2022-09-29
Observation Report Changes 936

johnatawnclementawn:
Add physical things that are related to an observation and relocate most of the content that was in the Parameters and Outcomes tab to the Description tab. This addresses #936 sans the spot map requested for the Related Objects tab. 

@njkim or @robgaston could either one of you review this please?

### DIFF
--- a/afs/media/js/views/components/reports/observation.js
+++ b/afs/media/js/views/components/reports/observation.js
@@ -14,13 +14,14 @@ define([
             params.configKeys = ['tabs', 'activeTabIndex'];
             Object.assign(self, reportUtils);
             self.sections = [
-                {id: 'name', title: 'Names and Classifications'},
-                {id: 'substance', title: 'Substance'},
-                {id: 'location', title: 'Location'},
-                {id: 'parameters', title: 'Parameters & Outcomes'},
-                {id: 'parthood', title: 'Parthood'},
+                {id: 'name', title: 'Names, Identifiers, Classification'},
                 {id: 'description', title: 'Description'},
+                {id: 'substance', title: 'Dates'},
+                {id: 'parameters', title: 'Parameters & Outcomes'},
+                {id: 'parthood', title: 'Parent Project'},
+                {id: 'objects', title: 'Related Objects'},
                 {id: 'documentation', title: 'Documentation'},
+                {id: 'data', title: 'Data'},
                 {id: 'json', title: 'JSON'},
             ];
             self.reportMetadata = ko.observable(params.report?.report_json);
--- a/afs/media/js/views/components/reports/observation.js
+++ b/afs/media/js/views/components/reports/observation.js
@@ -42,7 +42,6 @@ define([
             self.descriptionCards = {};
             self.documentationCards = {};
             self.substanceCards = {};
-            self.relatedObjectsCards = {};
             self.summary = params.summary;
 
             self.relatedObjectsTableConfig = {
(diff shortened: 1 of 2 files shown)


afs/media/js/views/components/reports/observation.js
```diff
@@ -132,11 +159,6 @@ define([
                                 value: self.getRawNodeValue(self.resource(), 'observed'),
```
njkim, line 159: The "object observed during observation" appears twice here and also in the related object tab. I suspect that this should be deleted.

afs/templates/views/components/reports/observation.htm
```diff
@@ -58,16 +64,7 @@ <h1 class="afs-report-toolbar-title"><span class="afs-report-name">{% trans "Obs
                     }
                 }"></div>
             </div>            
-            <!-- Location Tab -->
-            <div class="afs-report-page" data-bind="if: activeSection() === 'location'">
```
njkim, line 62: "Location" tab is removed by the ticket but it is still a visible node. I think it should appear somewhere. The description may be the place to go?

njkim (changes requested):
It looks great overall. Just a couple of minor comments.
1. The related object appears twice
2. The location does NOT appear at all in the report.

afs/media/js/views/components/reports/observation.js
```diff
@@ -44,6 +44,32 @@ define([
...
+            };
+
+            const resourceId = ko.unwrap(self.reportMetadata).resourceinstanceid;
+            const loadRelatedResources = async() => {
+                const result = await reportUtils.getRelatedResources(resourceId);
+                const physicalThingGraphID = '9519cb4f-b25b-11e9-8c7b-a4d18cec433a';
+                const relatedPhysicalThings = result?.related_resources.filter(resource => 
+                    resource?.graph_id === physicalThingGraphID);
+                
+                self.collectionOfRelatedObjects(relatedPhysicalThings.map(element => {
```
njkim, line 64: The link in the related-object is not working. getResourceLink is not going to work this way.
Also, this may be intentional but you can create a new object rather than adding to itself.

afs/templates/views/components/reports/observation.htm
```diff
@@ -76,7 +73,52 @@ <h1 class="afs-report-toolbar-title"><span class="afs-report-name">{% trans "Obs
...
+                                <div class="afs-table pad-btm">
+                                    <table class="table table-striped" cellspacing="0" width="100%">
+                                        <thead>
+                                            <tr class="afs-table-header">
+                                                <th>{% trans "Object" %}</th>
+                                                <th class="min-tabletl">{% trans "Description" %}</th>
+                                                <th class="afs-table-control all"></th>
+                                            </tr>
+                                        </thead>
+                                        <tbody data-bind="dataTablesForEach: { data: collectionOfRelatedObjects, dataTableOptions: relatedObjectsTableConfig }">
```
njkim, line 100: This table does not seem to follow the typical patterns... as in 
https://github.com/archesproject/arches-for-science-prj/blob/dev/6.2.x/afs/templates/views/components/reports/scenes/documentation.htm#L56-L71

johnatawnclementawn: Hi @njkim! I'm following the pattern seen in [project.htm](https://github.com/archesproject/arches-for-science-prj/blob/dev/6.2.x/afs/templates/views/components/reports/project.htm#L99-L111). Granted, I wrote that. I consulted with @aarongundel when diverging from that pattern.

njkim (approved):
Looks good!

Issue #1000, opened 2022-09-29
Create spot map in Observation report


aarongundel:
In the AFS workflows, we currently show a "spot map" of analysis areas.  This should be included in the observation report under the "Related Objects" tab.

aarongundel:
Per conversation between @njkim and @johnatawnclementawn - 

The general outline would be as follows: From the observation, you need to access the related dataset, which is held in the report. However, the sampled physical thing is what is shown in the spot map, so you need to getRelatedResources on the dataset and filter for physical thing (perhaps just filter by graphId). Then, you need to figure out which is the parent sample physical thing because under that node is where the annotations are held. (I believe Namjun said something about this node having a geometry, whereas the child sample nodes will not (?)). Then, perhaps once you have the data packaged up, you can call the annotation-summary ko component and render the spot map(?)

PR #1011, opened 2022-09-30
Add annotation to observation report, #1000

njkim:
Add annotation (spot map) of the area observed to the observation custom report, #1000 
In related object tab.


aarongundel (approved):
works for me!

Issue #1003, opened 2022-09-29
Update Search Result Details for Group


aarongundel:
The following should be present in the search result details for the group graph.

- Name 
- Name Type
- Parent group 
- Residence of associated location
- Professional activity type
- Identifier 
- Identifier type

PR #1013, opened 2022-10-04
The summary report for group resource, #1003 

njkim:
Update the summary report of the Group resource in the search result details, #1003 
Also, add a utility function to get the resource instance list value with the displayValue and the link.


aarongundel (approved):
looks good!

Issue #1004, opened 2022-09-29
Update Search Result Details for Place


aarongundel:
Update the search result details for place - the following should be present

- Name 
- Geospatial definition

PR #1015, opened 2022-10-04
The Place resource summary report, #1004

njkim:
Fix the location tab and update the summary report contents for the Place resource, #1004 
Update location scene to option to return only map.


aarongundel (approved):
looks good!

Issue #1010, opened 2022-09-29
Update search details for observation


aarongundel:
Update the search details for the observation graph to the following.

- Name
- Statement about observation
- Object observed
- Procedure/method
- Person carrying out
- Time span of observation

PR #1017, opened 2022-10-05
The observation summary report, #1010

njkim:
Update the summary report of the Observation resource, #1010


aarongundel (changes requested):
@njkim IMO these rows in these tables shouldn't hover as the mouse pointer cursor - since they don't go anywhere.  Should just be a minor style change.  Other than that, this is good to go.

njkim:
> @njkim IMO these rows in these tables shouldn't hover as the mouse pointer cursor - since they don't go anywhere. Should just be a minor style change. Other than that, this is good to go.

The pointer was removed in other pr. I hope that will resolve the pointer issue in general.
https://github.com/archesproject/arches-for-science-prj/pull/1013

aarongundel (approved):

PR #1018, opened 2022-10-05
dataset upload step changes

aarongundel:
This PR changes how the upload dataset workflow works.  
1.  If you want to add more files to a dataset (even with a different renderer) you can do that.  If the files in the archive share the same name, they will be replaced.  So you can repeatedly upload the same archive if things change.  
2.  Previous instructions on zip files can be disregarded.


aarongundel:
This PR changes how the upload dataset workflow works.  
1. It now requires a zip file containing all of the files that need to be uploaded to the dataset.  Do not nest files in other directories. 
2. A sub extension will determine the renderer that is used for the text files.  The available sub extensions are `.xrf`, `.raman`, and `.artax`.  The archive name should be as follows:  `filename.[subextension].zip`
3. If you want to add more files to a dataset (even with a different renderer) you can do that.  If the files in the archive share the same name, they will be replaced.  So you can repeatedly upload the same archive if things change.

aarongundel:
@njkim please verify the model is working as it should.  I think it's correct, but I need another set of eyes on it.

njkim:
@aarongundel 
It looks good. Working and saving as expected.
A couple of minor things are
1. I think we do not need __init__.py any more (there are two)?
2. There is not any way to change the dataset name (unless if is intentional)

And a few suggestions. (probably ok as they are)
1. The button says `Show Sampled Area List`. I wonder if it should be `Analysis Area` (I may be wrong about this)
2. The click on the annotation does not show the area in the left panel nor show any useful popup.
3. Once the dataset is created, even when all the files are removed the dataset is still there.

aarongundel:
@njkim regarding your previous comments - I think we do need __init__.py per the python module spec in order to do imports on those files correctly.  

The dataset name should be more clearly updatable now.  

The sampled area list is more than just analysis areas - there is another type in there.  

The annotation map should be showing popups now.  

There was a bug with the observation still remaining after delete; that is a separate ticket that has been merged in here now.

njkim:
It looks great and works as expected.
Just a couple of things.

1. The `save` button next to the dataset name does not seem to working for me.
I got:
`Tile update failed TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
It seems it is looking for files ($parent.saveFiles()). The files seem to be already saved upon uploading.
2. The popups on the annotation geometry all seem to say the parent physical thing. It would be great, if it shows the area resource? or maybe selecting the area in the table?
3. The message `Select a dataset and file to set and view interpretation and parameters.` in File Interpretation step initially partially hidden by the dataset details tab.

Thanks

aarongundel:
Hey @njkim thanks for the feedback - I've addressed all but your #1 suggestion.  I don't know how to easily address that as I think it's part of the IIIF functionality.  I think it would be good to address it though.  We can add that to the backlog as another issue to address in a different cycle.

njkim (approved):
It works great. Looks great too.
Thanks.

Issue #1008, opened 2022-09-29
Update search result details for sampling activity


aarongundel:
Update the search result details for sampling activity to the following 

- Parent Project of Sampling Activity
- Name
- [Form/type of sample: swab, cross-section, powder, scraping, mounted, unmounted material/chunk] – not currently in model, unless this is what’s supposed to be in the ‘type of sampling activity’ dropdown
- Statement

PR #1019, opened 2022-10-05
The sampling activity summary report, #1008

njkim:
Update the summary report of the sampling activity, #1008 

The sampling technique is saved in the workflow but as a statement with the generic description type since it does not have its own concept value.
So, the technique is shown as a statement along with the sampling type value.


aarongundel (approved):
looks good!

Issue #1021, opened 2022-10-06
Modify the invalid physical thing message in the upload dataset step


aarongundel:
Modify the invalid physical thing message in the upload dataset step.  As it is, the user cannot pick things in the first option that do not have an annotation.  For the second, users cannot pick areas.  Let the user know why the physical thing can't be used.

PR #1022, opened 2022-10-06
Add message in select object step, #1021

njkim:
Add the error messages when the invalid object is selected.

### DIFF
--- a/afs/templates/views/components/workflows/select-phys-thing-step.htm
+++ b/afs/templates/views/components/workflows/select-phys-thing-step.htm
@@ -16,7 +16,14 @@
     </div>
     <!--ko if: termFilter() -->
     <div style="padding: 5px; margin-top: 15px;">
-        <div class="widget-input-label">Analyzed Object or Sample</div>
+        <div class="widget-input-label">
+            <span data-bind="if: datasetRoute == 'non-destructive'">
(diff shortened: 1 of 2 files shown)


afs/templates/views/components/workflows/select-phys-thing-step.htm
```diff
@@ -35,7 +42,12 @@
     <!-- /ko -->
     <!-- ko if: validateThing && isPhysicalThingValid() === false -->
     <div class="tabbed-workflow-step-information-box widget-input-label" style="text-align: center; margin-top: 16px; height:fit-content; background-color: #fde9ad;">
-        The selected physical thing is invalid.  Complete the Analysis Area or Sample Taking workflow before proceeding.
+        <span data-bind="if: datasetRoute == 'non-destructive'">
+            You have to select an object with the analysis areas. Complete the Analysis Area workflow before proceeding.
```
aarongundel, line 46: Perhaps it might also be good to point them back to step one - something like  "this dataset type requires an object with an analysis area; please either complete the analysis area workflow or start the workflow over and choose 
 the sample dataset type"

aarongundel (changes requested):
looks good; minor wording tweak

aarongundel (approved):

Issue #1005, opened 2022-09-29
Update Search Result Details for Physical Thing


aarongundel:
The search result details for physical thing should be updated to the following:

- Name and name type as table
- Image of object (thumbnail from manifest) – not sure that this is currently in the resource model other than as linked to the ‘parts of object’. But would be useful.
- Statement
- Type of object
- Identifier and identifier type as table
- Current owner 
- Dimensions and type as table 
- Creator, date, Production event type, technique, location as a table 
- External URI

PR #1023, opened 2022-10-07
Update physical thing search details #1005

johnatawnclementawn:
Update the summary report of a physical thing #1005

### DIFF
--- a/afs/media/js/views/components/reports/physical-thing.js
+++ b/afs/media/js/views/components/reports/physical-thing.js
@@ -29,6 +29,14 @@ define([
                 columns: Array(6).fill(null)
             };
 
+            self.getTableConfig = (numberOfColumn) => {
+                return {
+                    ...self.defaultTableConfig,
+                    columns: Array(numberOfColumn).fill(null),
+                    columnDefs: []
+                }
+            };
+
             self.annotationTableHeader = 
                 `<tr class="afs-table-header">
                     <th>Region Name</th>
@@ -43,7 +51,13 @@ define([
             self.resource = ko.observable(self.reportMetadata()?.resource);
             self.displayname = ko.observable(ko.unwrap(self.reportMetadata)?.displayname);
             self.activeSection = ko.observable('name');
-            self.visible = {parts: ko.observable(true)};
+            self.visible = {
+                parts: ko.observable(true),
+                names: ko.observable(true),
+                statements: ko.observable(true),
+                identifiers: ko.observable(true),
+                creation: ko.observable(true)
+            };
             self.selectedAnnotationTileId = ko.observable(null);
             self.nameCards = {};
             self.descriptionCards = {};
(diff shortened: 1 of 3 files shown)


afs/media/js/views/components/reports/physical-thing.js
```diff
@@ -371,6 +386,137 @@ define([
...
+                    const content = self.getNodeValue(x, 'statement_content');
+                    const language = self.getNodeValue(x, 'statement_language');
+                    const tileid = self.getTileId(x);
+                    return { type, content, language, tileid }
+                }));
+            };
+
+            const typeData = self.resource()?.type;
+            if (typeData) {
+                self.typeSummary(Array({
```
aarongundel, line 468: Two things here - you can use cheap array initialization (instead of using the array constructor - just put in brackets []).  Second, looks like you found an interesting edge case for getRawNodeValue in report.js - I'm guessing you called @display_value here because I'm guessing not doing so returned an error for you.  I think that's actually because of lines 31-33 in report.js - that they're in the array loop.  I suspect that's incorrect and it would be better to actually move those lines and line 25 outside of the loop.  Without testing it, that should fix this behavior and you won't need to specify @display_value after that.

afs/media/js/views/components/reports/physical-thing.js
```diff
@@ -371,6 +386,137 @@ define([
...
+                    const link = self.getResourceLink({resourceId: self.getNodeValue(x, 'resourceId')});
+                    return { displayValue, link }
+                }));
+            };
+
+            const dimensionData = self.resource()?.dimension;
+            if (dimensionData) {
+                self.dimensionsSummary(dimensionData.map(x => {
+                    const value = self.getNodeValue(x, 'Dimension_value ');
+                    const unit = self.getNodeValue(x?.dimension_unit, '@display_value');
```
aarongundel, line 496: you should be able to specify x, 'dimension_unit' here.  Avoid requesting @display_value directly.

afs/media/js/views/components/reports/physical-thing.js
```diff
@@ -371,6 +386,137 @@ define([
...
+                }));
+            };
+
+            const creationData = self.resource()?.production;
+            if (creationData) { 
+                self.creationSummary(Array({
+                    creator: self.getNodeValue(creationData, 'production_carried out by'),
+                    creationDate: self.getNodeValue(creationData?.production_time, 'production_time_begin of the begin'),
+                    type: self.getNodeValue(creationData, 'production_type'),
+                    technique: self.getNodeValue(creationData?.production_technique, '@display_value'),
```
aarongundel, line 508: same comments as above re: display_value

afs/templates/views/components/reports/physical-thing.htm
```diff
@@ -184,21 +184,240 @@ <h1 class="afs-summary-report-title"><span data-bind="text: displayname" class="
...
+                                <td data-bind="text: type"></td>
+                            </tr>
+                        </tbody>
+                    </table>
+                </div>
+                <!-- /ko -->
+            </div>
+        </div>
+        <!-- Image -->
+        <div class="afs-report-section">
```
aarongundel, line 214: Can you add a header here in the same font style as the table column headers?  There's clarity in what data is displayed in the others; but because this isn't a table, it's harder to know what exactly it is.  Something like "Image Manifests" may do.

aarongundel (changes requested):
Few minor changes.

afs/templates/views/components/reports/physical-thing.htm
```diff
@@ -184,21 +184,240 @@ <h1 class="afs-summary-report-title"><span data-bind="text: displayname" class="
...
+                        <img data-bind="attr: { src: thumbnail }">
+                    </span>
+                </div>
+                <!-- /ko -->
+            </div>
+        </div>
+        <!-- Statements -->
+        <div class="afs-report-section">
+            <div class="afs-report-subsection">
+                <!-- ko ifnot: statementsSummary() && statementsSummary().length -->
```
aarongundel, line 233: Also, can you remove these "No statements/images/names" lines for the time being..  Namjun isn't using these in his summaries, so it appears inconsistent.

aarongundel:
looks good!

Issue #963, opened 2022-09-08
Project title cut off in left side bar


johnatawnclementawn:
When the left side bar is expanded, the word 'Science' is cut off.

![Missing 'Science' in the project title](https://user-images.githubusercontent.com/22161487/189221176-e5d3cfc1-4107-4e5d-8f77-1163a8c7fa46.png)

chiatt:
Resolved in #1024

PR #1024, opened 2022-10-07
Show full title when navbar is expanded #963

johnatawnclementawn:
Stop the word 'Science' from being cut off when the navbar is expanded #963


njkim (approved):
Look good!

Issue #993, opened 2022-09-26
Sample Taking, Analysis Area workflows: need a way to delete an existing sample or analysis area


dwuthrich:
As shown on the accompanying image, users can create new sample areas or edit existing areas.  But there is no way to delete an existing area from the sample list widget.

What should happen:
Add a "delete" href (could just be a trashcan icon) to sample areas that are not locked.  Clicking on icon opens an alert box confirming that the user wants to delete the sample area (similar for the analysis area workflow).  ![Screen Shot 2022-09-26 at 2.18.33 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/766e9b31-439e-4288-aefb-6023ac3d26)

Users should be able to delete newly created analysis areas in the analysis areas workflow. **Users should not be able to delete analysis areas that are related to a digital resource because that would indicate that there is an observation already associated to the analysis area.**

Users should be able to delete newly created samples in the sample taking step of the sampling activity workflow. **Users should not be able to delete samples that are 'locked' because that indicates they are related to a prior sampling activity and therefore should not be modified.**

dwuthrich:
![Screen Shot 2022-09-26 at 2 18 33 PM](https://user-images.githubusercontent.com/815185/192383004-c040918c-7b40-49a9-9ca3-f79e59b63087.png)

PR #1027, opened 2022-10-10
993 delete sample

apeters:
allow deletion of samples and analysis areas re #993


njkim (approved):
Works great and look great!

Issue #1025, opened 2022-10-07
Sample Taking/Analysis Area Workflows: Allow user to select an existing manifest in Image step


dwuthrich:
The sample taking and Analysis area workflows require a user to associate a IIIF manifest with the object being sampled.  Currently the step only allows users to create a new manifest (as shown in the attached image).

What should happen:  a user should also have the option to select an exiting manifest in this step.  We should add a dropdown that lists existing manifests as the current Image Service Manager does.

![Screen Shot 2022-10-07 at 1.13.30 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/d4bd86ca-21e7-4239-a723-7df824f7d3f6)

PR #1028, opened 2022-10-11
Add capability to choose a existing image service during workflow, #1025 

njkim:
Add the capability to choose the existing manifest (image service) during the image service selection step in sample taking, analysis area workflows, #1025 
By utilizing the existing edit tab from the image service manager (previously disabled in the workflow)


aarongundel (approved):
works for me.

Issue #1026, opened 2022-10-07
Sample Taking Workflow: Summary step of workflow does not render


dwuthrich:
The last step of the Sample Taking worklfow does not show the summary information for the user to review before saving or rejecting the workflow:

![Screen Shot 2022-10-07 at 1.40.22 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/83f9b262-1957-43e2-ac99-3ae39896ac36)

PR #1031, opened 2022-10-13
Fix sampling final summary, #1026

njkim:
Fix the sample taking workflow final step, where the page does not render if there are more than one sample on the same annotation canvas, #1026


aarongundel (approved):
This works for me.

Issue #1034, opened 2022-10-18
Multiple observations created when moving through the upload dataset workflow


aarongundel:
If you go through the upload dataset workflow - pass the step where the new observation is created, and then go back to that step again.  The "Save and Continue" button lights up (it shouldn't, since nothing has changed).  If you press that button to move forward, the workflow advances and a second observation is created.  The workflow should only create one observation per run.

apeters:
I"m not able to replicate this.

aarongundel:
@apeters after looking at this, you need to refresh the page to replicate.  I think there's something that isn't saved in the workflow state properly.  So create the observation, go forward, then go back and refresh the page.

PR #1037, opened 2022-10-19
fix issue with duplicate observations being created, re #1034

apeters:


aarongundel (approved):
looks good and works for me!

Issue #1002, opened 2022-09-29
Update Search Result Details for Person


aarongundel:
For the Person type, these details should be visible.  

- Name
- Group person is a member of
- Contact information (if not blank); specifically email
- External URI (if one exists)
- Core attributes: Nationality only
- Birth date only 
- Death date only

PR #1038, opened 2022-10-19
Update Person search details #1002

johnatawnclementawn:
Update the summary report of a person #1002

### DIFF
--- a/afs/media/js/views/components/reports/person.js
+++ b/afs/media/js/views/components/reports/person.js
@@ -29,6 +29,13 @@ define([
                 ...self.defaultTableConfig,
                 columns: Array(3).fill(null)
             };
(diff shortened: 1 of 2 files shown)


afs/media/js/views/components/reports/person.js
```diff
@@ -246,6 +253,75 @@ define([
...
+            };
+
+            const nationalityData = self.resource()?.nationality;
+            if (nationalityData) {
+                self.nationalitySummary([{
+                    content: self.getNodeValue(nationalityData)
+                }]);
+            }
+
+            const birthData = self.resource()?.birth?.Birth_time;
```
aarongundel, line 306: FYI - you can get this info with getNodeValue.  Also, getNodeValue is case insensitive.  So you can do all lowercase names.

Issue #1042, opened 2022-10-24
Physical thing report doesn't display name


aarongundel:
The physical thing report does not display the name of the physical thing when it is provided.  
![Screen Shot 2022-10-24 at 3 26 59 PM](https://user-images.githubusercontent.com/124614/197632886-4a1ac0d0-8566-484f-8373-f9f87dd41c1c.png)

Also, while you're at it - please tweak the alignment of the plus button.

PR #1043, opened 2022-10-25
Fix names in custom reports, #1042

njkim:
Fix the issue of names are not showing in the custom reports of the multiple resource models, #1042
Also, fix the position of buttons


aarongundel (approved):
This works - thanks @njkim !

Issue #1065, opened 2023-01-06
Migrate Reports to Arches v7.2


chiatt:

chiatt:
Resolved in #1076

PR #1076, opened 2023-01-13
Format custom report templates to load in v7, re #1065

chiatt:
Modifies format or reports and scenes components so that reports at least load. More work is needed to get data to render properly in reports.


johnatawnclementawn:
Looks good

Issue #1058, opened 2023-01-06
Migrate Analysis Area Workflow to Arches v7.2


chiatt:

chiatt:
Resolved in #1085

PR #1085, opened 2023-01-20
Update the analysis area workflow for v. 7.x, #1058

njkim:
Update the analysis area workflow to work in arches 7.x, #1058

*** dependent on the core arches branch: afs_1062_tile_loading_issue


afs/views/analysis_area_and_sample_taking.py
```diff
@@ -173,6 +174,11 @@ def post(self, request):
         part_identifier_assignment_tile_data = JSONDeserializer().deserialize(request.POST.get("partIdentifierAssignmentTileData"))
         part_identifier_assignment_tile_id = request.POST.get("partIdentifierAssignmentTileId") or None
         name = request.POST.get("analysisAreaName")
+        print("++++++++++++++++++++++")
+        print(name)
+        print("++++++++++++++++++++++")
```
chiatt, line 179: Maybe these print statements can be removed

njkim: thanks

Issue #1063, opened 2023-01-06
Migrate Upload Dataset Workflow to Arches v7.2


chiatt:

chiatt:
Fixed in #1091

PR #1091, opened 2023-01-25
Upgrade the upload dataset workflow for v.7, #1063

njkim:
Upgrade the upload dataset workflow for v.7, #1063


chiatt (approved):
This looks good and is working for me, but I think all the 'buildStrObject' methods should be moved into a separate module that can be reused across steps.

Issue #1061, opened 2023-01-06
Migrate Review Dataset Workflow to Arches v7.2


chiatt:

chiatt:
Resolved in #1092

PR #1092, opened 2023-01-26
Upgrade review dataset workflow v.7, #1061

njkim:
Upgrade review dataset workflow v.7, #1061
Also add the dataset name in the file-list in the file interpretation step. #1098


chiatt (approved):
Looks good to me

Issue #1090, opened 2023-01-24
Canceling edits does not update analysis area name


chiatt:
Edits to 'Part Name' strings are not undone when a user clicks 'cancel' in the analysis area workflow. The edit is undone. This can be observed in the preview of the part name, but the change in the string widget (above the preview) persists. In the example below 'Lawn' should be 'Grass' after canceling the change from 'Grass' to 'Lawn'.

![Image](https://user-images.githubusercontent.com/1877663/214440314-21d62263-fc16-4151-9ad6-447533009067.png)

chiatt:
Fixed in #1101

PR #1101, opened 2023-02-01
Fix cancel function of text edit in workflows, #1090 

njkim:
Fix cancel function in annotation steps in analysis area & sample taking workflows, #1090


chiatt (approved):
This looks good and resolves the issue with the string widgets. There still appears to be an issue with canceling edits to existing geometries, but that could be addressed in a different ticket.

PR #1103, opened 2023-02-03
Fix indexing for tiles - defer to resource save

aarongundel:
Fixes related to deferred resource indexing


chiatt (approved):
Looks great!

Issue #1108, opened 2023-02-13
Update Project Workflow occasionally blows up


aarongundel:
It appears that the Update Project Workflow occasionally blows up when an empty array is provided to the getstringvalue function or the striptags function.  Optional chaining will fix this.

aarongundel:
Fixing this will fix the corresponding issue in the create new project workflow

chiatt:
Fixed in #1110

Issue #1096, opened 2023-01-26
summary tab within multi-object project is blank


dmaclennan4:
<img width="794" alt="issue3" src="https://user-images.githubusercontent.com/120419245/214967647-9eda4b6d-34e0-4b17-a043-c731c954306e.PNG">

summary tab is not populated for a multi-object project (Glass Project), but is populated for a single object project (2018.40 Rubens)

dmaclennan4:
Update - we made a "mini" test of 3 objects, and they populated in the summary tab

<img width="795" alt="issue3 1" src="https://user-images.githubusercontent.com/120419245/214968776-2b402ec2-4a71-4179-a037-721f1e22e430.PNG">

chiatt:
Addressed in #1110

PR #1110, opened 2023-02-14
1108 fix update project

aarongundel:
Fixes for #1108 and #1096 - fixes some workflow issues that happen when the description (or other string based i18n value) is null/undefined - and fixes an issue that occurs in the base workflow when there is no save state for a given step by making said (add things) step required in the update workflow.


chiatt (approved):
Looks good!

PR #1127, opened 2023-03-13
chemical image workflow

aarongundel:
chemical analysis workflow first attempt


aarongundel:
> It looks like boto needs to be added to the project requiremnts.txt file `ModuleNotFoundError: No module named 'boto3'`

It's already there - see https://github.com/archesproject/arches-for-science-prj/blob/dev/7.4.x/afs/install/requirements.txt

chiatt (changes requested):
This looks great - just a few minor changes before we merge it in:

1. Change workflow name to Chemical Image Upload.
2. Crud buttons should not appear below Dataset Name input if user uploads a file before entering a name
3. Workflow should only be available to users that have cloud storage configured
4. The word manifest should in the init-workflow page should be changed to 'image service'

Issue #1129, opened 2023-03-28
Collection name does not appear in summary step of new project wf


chiatt:
The related collection name does appears as Name_content in the summary step of new project wf:

![Image](https://user-images.githubusercontent.com/1877663/228355384-d0c6236d-b1bc-4624-8777-77179ecee982.png)

PR #1130, opened 2023-03-30
Add descriptor to the add things step backend, #1129

njkim:
Add descriptor before indexing for the add things step (new project wf), #1129


chiatt (approved):
Looks good - fixes #1129

Issue #1128, opened 2023-03-23
Dirty state activated after saving a sample


chiatt:
In the sample taking workflow, after a user saves a sample in the sample taking step, the dirty state remains activated. This makes it a little unclear to the user that the sample has indeed been saved.

PR #1131, opened 2023-03-30
Fix the dirty state for the sampling location step, #1128

njkim:
Fix the dirty state to show not-dirty properly after saving the tiles.

### DIFF
--- a/afs/media/js/views/components/workflows/sample-taking-workflow/sample-taking-sample-location-step.js
+++ b/afs/media/js/views/components/workflows/sample-taking-workflow/sample-taking-sample-location-step.js
@@ -660,7 +660,7 @@ define([
                     cardwidgets: self.card.params.cardwidgets,
                 });
 
-                self.selectedSampleLocationInstance(self.builtTile);
+                self.selectSampleLocationInstance(self.builtTile);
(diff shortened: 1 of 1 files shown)


afs/media/js/views/components/workflows/sample-taking-workflow/sample-taking-sample-location-step.js
```diff
@@ -657,7 +660,7 @@ define([
                     cardwidgets: self.card.params.cardwidgets,
                 });
 
-                self.selectedSampleLocationInstance(self.builtTile);
+                self.selectSampleLocationInstance(self.builtTile);
```
chiatt, line 663: This change looks like it introduces a bug. If you create a sample with a polygon, save, then create another sample with a point, you'll notice that you can't see the point annotation unless you refresh the page. Removing this change, the dirty state issue looks resolved by the changes on lines 104 and 111.

chiatt (changes requested):
Changes on 104 and 111 seem to correct the problem, but the change on 660 seems to introduce a new bug.

afs/media/js/views/components/workflows/sample-taking-workflow/sample-taking-sample-location-step.js
njkim: interesting... updated.

chiatt (approved):

Issue #973, opened 2022-09-14
Handling image services not related to physical thing


chiatt:
During the analysis area and sample taking workflows, a user has the option to create a new image service. Once a user creates a service they can see that service listed in the step. If they save and proceed to the next workflow step, the image service is gets related to the physical thing it represents. However, if the user quits the workflow at this point, the image service remains in the system, but no relationship is saved to the physical thing. Unfortunately, because there is no relationship saved, the image service will not appear in this step if a user goes through the workflow again. The next time the user goes through the workflow, if they need an image service for their physical thing, they have to create a new one. If they do that, then there are two image services for one physical thing. This could cause problems over time if users do any work in the resource editor where they could create relationships to the wrong image service if there are orphaned image services for a given physical thing.

To handle for this we could:
1. Ensure that the image service is deleted if a user deletes the workflow (this doesn't help a user simply closes the tab)
2. Allow users to either create a new image service or select from existing image services.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/64cc41ca-79fe-4cb1-b492-3347f12b5c76
)

apeters:
I don't actually observe this behaviour.  If I go to the next step after creating an image service and then delete the workflow.  The image service is deleted and so is the relationship.  Maybe I'm missing something?

apeters:
@aarongundel since Cyrus is out, could you confirm that this is still an issue

njkim:
@apeters
Does the image service record itself is deleted?
Since the workflow only shows the related image services, it may look like the image service is deleted.
But it is possible that the relationship is deleted but the image service isn't.

chiatt:
It looks like this is still an issue. The relationship is deleted, but the image service record (the record in the iiif_manifests table, and related images) still persist.

PR #1132, opened 2023-04-03
Add transaction id to manifest manager call, #973

njkim:
Add transaction id to manifest so that it can be removed when the workflows is cancelled, #973 

Depends on the https://github.com/archesproject/arches/pull/9396


chiatt (approved):
Looks good!

Issue #1075, opened 2023-01-13
Textual Work is using a tabbed template rather than a custom template


chiatt:
This will require a graph change so that textual work is using the custom template rather than the default tabbed template.

![Image](https://user-images.githubusercontent.com/1877663/212366536-aece50ac-33fc-40c5-b01e-77a08b5c6c66.png)

PR #1146, opened 2023-05-01
Link the textual work report template, #1075

njkim:
Update textual work models to link the proper report template, #1075


chiatt (approved):
Looks good!

Issue #1145, opened 2023-05-01
Improve loading bar UI of the Chemical Analysis Data workflow


chiatt:
Currently the loading indicator runs along the top of the file input. It should be made more prominent and outside of the file input.

PR #1150, opened 2023-05-03
Move and style chemical analysis file upload progress bar, re #1145

ryan86:


chiatt (changes requested):
Hi Ryan - this looks good. I've suggested just a couple of tweaks to tighten it up a bit.

chiatt (approved):

Issue #1136, opened 2023-04-12
Add step in reporting workflow download related files


chiatt:

chiatt:
Users will need to download files associated with physical things in a report. A workflow step should list all of the physical things, their digital resources, and each digital resources's files. Users should then be able to check each of the files that they want to download. A notification should then be sent to the users with a link to files larger than 4gb and a link to a zip file with files smaller than 4gb.

PR #1152, opened 2023-05-04
Download files step in report wf, #1136

njkim:
Add a step to project report workflow that allow the user to download the files that are related to the project on interest.

This pr is dependent to the core arches pr, https://github.com/archesproject/arches/pull/9505


afs/media/js/views/components/workflows/project-report-workflow/download-project-files.js
```diff
@@ -0,0 +1,141 @@
...
+                return !!observation.relatedFiles().find(file => file.selected() == true );
+            });
+        });
+
+        this.formatSize = function(size) {
+            var bytes = size;
+            if(bytes == 0) return '0 Byte';
+            var k = 1024;
+            var dm = 2;
+            var sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
```
aarongundel, line 33: It's not clear from a simple reading what this function does (I think you're making the file size human-readable).  Some of these sizes are unnecessary - I doubt we'd ever have terabyte sized files, much less petabyte level files.  Some additional comments and renaming of the variables in this function would be helpful - as well as making these vars const.

afs/media/js/views/components/workflows/project-report-workflow/download-project-files.js
```diff
@@ -0,0 +1,141 @@
...
+        this.selectAll = function(bool) {
+            self.relatedObservations().forEach((observation) => {
+                observation.relatedFiles().forEach(file => file.selected(bool));
+            });
+            self.expandAll(true);
+        };
+
+        this.getFilesFromObservation = async() => {
+            let projectObservations, observation, digitalResources;
+            await window.fetch(arches.urls.related_resources + self.projectValue  + "?paginate=false")
```
aarongundel, line 57: If you're using an await in an async context, there is no need to chain a "then".  You can simply "await" it as well and remove the chaining to make this easier to read.  For instance, instead of saying `.then(response=>response.json())` you can simply say `const response = await response.json();`.  This will cut down on your nested functions.

aarongundel (changes requested):
Mostly text tweaks and some minor code changes - looks pretty good!

aarongundel (changes requested):
Mostly text tweaks and some minor code changes - looks pretty good!

afs/media/js/views/components/workflows/project-report-workflow/download-project-files.js
njkim: It is not used any more so removed.

Issue #1148, opened 2023-05-01
Render manifest manager in workflow step over top bar (bar with login/user profile btns)


chiatt:
Currently when the manifest manager appears in a workflow, the 'Return to Workflow' button floats awkwardly above the sidebar header. Instead he manifest manager should take up the entire page to the right of the main nav bar so that there is a place to put the 'Return to Workflow' button.

![Image](https://user-images.githubusercontent.com/1877663/235529383-71d9d4b0-a553-4d0b-82cd-1ba573b4a081.png)

PR #1153, opened 2023-05-06
1148 manifest manager top bar

ryan86:
Renders manifest manager top bar on top of Arches top bar re #1148 .


chiatt (changes requested):
This looks really good @ryan86. Go ahead and apply this to the other workflows that load the manifest manager. Because they all use the same css, could you centralize the css in the project.css file?

chiatt (approved):
Looks good to me!

Issue #1163, opened 2023-05-11
Add 'Select Template' step to Project Report Workflow


johnatawnclementawn:

johnatawnclementawn:
Resolved in #1165

PR #1165, opened 2023-05-15
1163 Add select template step to project report WF

johnatawnclementawn:
Moves the report template selection from the initial project report workflow step to its own step & displays available templates not in a select dropdown but with a radio button & corresponding preview thumbnail. 

Addresses: #1163


afs/media/js/views/components/workflows/project-report-workflow/report-template-select.js
```diff
@@ -0,0 +1,69 @@
...
+    'knockout',
+    'knockout-mapping',
+    'templates/views/components/workflows/project-report-workflow/report-template-select.htm'
+], function(_, arches, ko, koMapping, reportTemplateSelectTemplate) {
+    function viewModel(params) {
+        this.templateValue = ko.observable();
+
+        let archesTemplates = [];
+        this.allTemplateFormats = ko.observableArray();
+        this.pptTemplates = ko.observableArray();
```
aarongundel, line 13: Seems like these could be just arrays instead of observable Arrays?

aarongundel (changes requested):
This looks good - just a minor code tweak.  I think we need to allow for the user _not_ to include a thumbnail as well - in which you'd just render some generic image for all word docs or all powerpoints.  Additionally, if there is a preview provided (which will be a pdf) - we need the image to link to that preview and open it on the same page - without leaving the current workflow step, so they can view what's in the preview.  If that preview isn't provided, the image shouldn't be a link.

afs/media/js/views/components/workflows/project-report-workflow/report-template-select.js
```diff
@@ -0,0 +1,82 @@
...
+                        templates: this.pptTemplates,
+                        icon: 'fa fa-file-powerpoint-o',
+                    },
+                    // {
+                    //     heading: 'Excel Templates',
+                    //     templates: this.xlsxTemplates,
+                    //     icon: 'fa fa-file-excel-o',
+                    // },
+                ]);
+            } catch (error) {
```
aarongundel, line 63: Are you trying to catch and recover from a specific error here?  These kinds of things can be frustrating to debug unless you rethrow - or just don't catch the error.

aarongundel (approved):

aarongundel (approved):
LGTM

PR #1172, opened 2023-05-22
removes iiif manifests; used in getty's afs package

aarongundel:
removes iiif manifests from afs base package


chiatt (approved):
Looks good!

PR #1173, opened 2023-05-23
1166 Improve download report step

johnatawnclementawn:
Makes things look nice and gives user more information about report that will be downloaded

### DIFF
--- a/afs/templates/views/components/workflows/project-report-workflow/report-template-select.htm
+++ b/afs/templates/views/components/workflows/project-report-workflow/report-template-select.htm
@@ -37,7 +37,7 @@
                         $parent.showPDFPreview(preview);
                     }, 
                     attr: {
-                        'aria-expanded': $parent.previewSelected(),
+                        'aria-expanded': $parent.previewSelected().toString(),
                         title: 'Preview ' + name + ' report'
                     }"
                 >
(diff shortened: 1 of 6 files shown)


afs/templates/views/components/workflows/project-report-workflow/report-template-select.htm
```diff
@@ -62,11 +62,9 @@
                 data-bind="css: { 
                         'active': $parent.templateValue() === id 
                     },
-                    click: function () {
```
aarongundel, line 65: Why did this get changed?  This was more correct the first time (wrapping it in anonymous function - which ensures the function is not called on initialization; you assign a function vs calling that function) - if you don't do this, you'll get a default value set for the template - which may have been your intention, but if you do it like this, template name doesn't get set - and you get [Object object] in the "summary/download" step.  Try this with a single template in your db and you'll see what I mean.

aarongundel (changes requested):
Mostly ok; one issue with a single template workflow.  Also can you purge the "sectionTitle"s from project-report-workflow.js?

aarongundel (changes requested):
Mostly ok; one issue with a single template workflow.  Also can you purge the "sectionTitle"s from project-report-workflow.js?

afs/templates/views/components/workflows/project-report-workflow/report-template-select.htm
johnatawnclementawn: This was an artifact of me trying to debug passing values between workflow steps. I think everything should be sorted now.

aarongundel (approved):
LGTM

PR #1174, opened 2023-05-24
add depicts node to Physical Thing

annabelleee:
Added depicts node to Physical Thing model, but I notice there are other changes in the model that I did not make. Please check to make sure this is okay since I used the model version on https://afs.test.fargeo.com/.

### DIFF
--- a/afs/pkg/graphs/resource_models/Physical Thing.json
+++ b/afs/pkg/graphs/resource_models/Physical Thing.json
@@ -264,33 +264,6 @@
                 },
                 {
                     "active": true,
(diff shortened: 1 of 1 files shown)


chiatt:
Thanks @annabelleee, I'll have a look

chiatt:
@annabelleee I think a lot of the changes can be explained by some recent database migrations, but it looks like there might be an unintended node, `New_Node_1`. 
image

annabelleee:
You're right. I didn't see that and I thought I deleted all the extra nodes. Should I close this and create a new PR?

annabelleee:
@chiatt See above. I just pushed the commit where I removed that extra node.

Issue #1139, opened 2023-04-13
Sample Taking Workflow null name


aarongundel:
There's an issue in the sample taking workflow.  If you create a new sample, after you've done so, click the external green "Save" button at the bottom.  The last sample in the list will have its name turn "null".   The real problem here is that after you save the sample, the save button is still green - it should be disabled.

chiatt:
Dang - I thought @njkim had this resolved. https://github.com/archesproject/arches-for-science-prj/pull/1131

PR #1187, opened 2023-06-12
Refresh the newly created sample locations, #1139

njkim:
Fix the issue with dirty state isn't working properly which results in the unexpected behavior such as sample name becomes null, #1139


chiatt (approved):
Looks good!

PR #1193, opened 2023-06-19
removes create manifest step from CA workflow

aarongundel:
Removes create manifest step from CA/data upload workflow


chiatt (approved):
Looks good!

PR #1201, opened 2023-06-21
improves error message in select physical thing step

aarongundel:
improves error message in select physical thing step

### DIFF
--- a/afs/templates/views/components/workflows/select-phys-thing-step.htm
+++ b/afs/templates/views/components/workflows/select-phys-thing-step.htm
@@ -49,7 +49,7 @@
             The selected object must have at least one related sample location or analysis area.
         </span>
         <span data-bind="if: datasetRoute == 'destructive'">
-            You have to select a sample. Please complete the sample taking workflow or start the workflow over and select the area dataset type.
+            The selected object must be a sample.
         </span>
     </div>
     <!-- /ko -->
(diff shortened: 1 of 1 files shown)


afs/templates/views/components/workflows/select-phys-thing-step.htm
```diff
@@ -46,7 +46,7 @@
     <!-- ko if: validateThing && isPhysicalThingValid() === false -->
     <div class="tabbed-workflow-step-information-box widget-input-label" style="text-align: center; margin-top: 16px; height:fit-content; background-color: #fde9ad;">
         <span data-bind="if: datasetRoute == 'non-destructive'">
-            You have to select an object with the analysis areas. Please complete the analysis area workflow or start the workflow over and select the sample dataset type.
+            The selected object must have at least one related sample location or analysis area.  Please complete the analysis area workflow or start the workflow over and select the sample dataset type.
```
chiatt, line 49: This message still seems unclear, particularly the second half. What if the user isn't selecting a sample here? If they start the workflow over and get to this step, and reselect the object, they are just going to see a different error message which will make things even more unclear. Also, there are other ways they can solve this problem. They could just pick a different physical thing, or they could through the sample taking workflow and take a sample which would create an analysis area. 


The best solution would be to filter the options for only qualified objects. Until we can do that, I think the message should actually be simple, like: ```The selected object must have at least one related sample location or analysis area.```

chiatt (changes requested):
I think line 52 could be improved as well. Maybe it should be: ```The selected object must be a sample.```

chiatt (approved):

PR #1203, opened 2023-06-26
fixes issue with datasetname in uploadxy workflow

### LINKED ISSUE #1185: Upload XY Data Workflow Bugs: Labels and adding new files (dev/7.4.x branch)
I found a bug in the dev/7.4.x branch Upload XY Data Workflow

In the "Manage Dataset" interface, there's a bug in rendering the Dataset Names. The two text boxes read [object Object], perhaps because they expect simple strings and not language indicating dictionary objects?

Also, I can't seem to add more files to Analysis Area that I have selected. This is the second time I'm trying to use this workflow to add image files to an Analysis Area (physical thing), but in this second use of the workflow, I'm unable to add additional files. 

Finally, and this may be another issue, but this seems to be the only way I can add image files to associate with an Analysis Area. These JPEG files are rasters, but they don't seem that appropriate to an "XY" data workflow (which seems to expect data in some sort of structured text format), but I lack any other workflow options for associating additional images to Analysis Areas. 

afs-xy-workflow-cannot-add-new-file-label-issue

aarongundel:
fixes issue with datasetname in uploadxy workflow

closes #1185

### DIFF
--- a/afs/media/js/views/components/workflows/upload-dataset/select-dataset-files-step.js
+++ b/afs/media/js/views/components/workflows/upload-dataset/select-dataset-files-step.js
@@ -24,6 +24,10 @@ define([
                 }};
             };
 
+            this.geti18nStringValue = i18nObj => {
+                return i18nObj?.[arches.activeLanguage]?.value || (typeof i18nObj === 'object' && i18nObj !== null ? '' : i18nObj);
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Looks good!

PR #1207, opened 2023-06-28
remove user from request to load_tiles

### LINKED ISSUE #1206: non-admin user cannot advance through upload dataset workflow
A non-admin user cannot advance through the upload dataset workflow; no permissions are provided in the call to load tiles.

aarongundel:
user was not needed.  resolves #1206


chiatt (approved):
Looks good!

Issue #1192, opened 2023-06-19
Add version number to project


chiatt:
Version should indicate that the 7.4.x branch we are working on 1.0 beta

PR #1208, opened 2023-06-28
Add project version to AfS, re #1192

chiatt:


aarongundel (approved):
This is good, I don't see any issues here right now - but it seems a little onerous to have to increment the max arches version every time there's a point release.  This will probably cause some friction when doing new releases of core in response to security issues and such.  It seems like we'd want to allow all point releases of a particular version like 7.4.  In that case it seems like we'd need to favor a range of versions.  https://python-semanticversion.readthedocs.io/en/latest/guide.html#range-specifications

PR #1212, opened 2023-06-30
1195 save button fix

### LINKED ISSUE #1195: Save button remains enabled after dataset is saved
It should be disabled after save, or there should be some other indication that the dataset has been successfully saved

image

aarongundel:
Fixes #1195 by causing the save button to disappear if the name of the dataset is not dirty.


chiatt (changes requested):
When I try to change the name, an error claims self.nameDirty is undefined

image

chiatt (approved):
Looks good!

PR #1214, opened 2023-07-06
unchecked responses fail with mysterious messages

aarongundel:
Some fetch requests aren't being properly checked for response codes.  These can cause strange error messages and throw the end user off - making them think there's a frontend issue or somesuch, rather than a bad response from the backend.


chiatt (approved):
Looks good!

Issue #1199, opened 2023-06-20
Download Selected button should be blue rather than red


chiatt:
<img width="658" alt="image" src="https://github.com/archesproject/arches-for-science-prj/assets/1877663/e2465465-c531-4b52-9afd-ced843e0d0e3">

PR #1220, opened 2023-07-13
Update download button color to blue, #1199

njkim:
Change download button color in create report workflow, re #1199


chiatt (approved):
Looks good!

PR #1230, opened 2023-07-25
Update rdm, models, labels

annabelleee:
Includes:
New Observation Types
New Digital Resource Types
New File Types under Digital Resources
New Relationship Labels (arches thesaurus, new collections, model changes)


annabelleee:
@chiatt There were some nodes that I could not change the property labels for because they were frozen (regardless of whether I unpublished or dumped the resources)... it seems that the nodes that were frozen participate in the workflows? Please verify and let me know the best way to add the property labels to those.

chiatt (approved):
Looks good to me! Tested in all workflows without any issue.

Issue #1235, opened 2023-07-30
Ensure relationship types are saved in new project workflow


chiatt:
Ensure that the proper forward and inverse relationship types are saved in new project workflow

PR #1236, opened 2023-07-30
Add relation types for collections, re #1235

chiatt:
Ensures relationship types are defined from a project to its collection and from a thing to its collection. re #1285


aarongundel (approved):
LGTM

Issue #1237, opened 2023-07-30
Ensure Relationship Types are Saved in Analysis Areas Workflow


chiatt:
Ensure forward and inverse relationship types are saved when going through the Analysis Areas workflow

PR #1238, opened 2023-07-30
Saves analysis area relationship types, re #1237

chiatt:
This applies to some relationship types in sample taking as well, but not all.

### DIFF
--- a/afs/views/analysis_area_and_sample_taking.py
+++ b/afs/views/analysis_area_and_sample_taking.py
@@ -16,12 +16,12 @@ from arches.app.utils.response import JSONResponse
 logger = logging.getLogger(__name__)
 
 
-def get_related_resource_template(resourceid):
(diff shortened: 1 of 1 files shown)


jacobtylerwalls (commented):
I noticed there are still some null relationship types being created for this workflow here, b/c they are not provided to `get_related_resource_template`:

https://github.com/archesproject/arches-for-science-prj/blob/600c7e0516ff894008826ef78297e7701e53040f/afs/views/analysis_area_and_sample_taking.py#L308-L311

Is that expected?

njkim:
Looks like this will be resolved in this ticket, #1243

> I noticed there are still some null relationship types being created for this workflow here, b/c they are not provided to `get_related_resource_template`:
> 
> Is that expected?

njkim (approved):
Looks good to me

PR #1239, opened 2023-08-01
Avoid GET requests when selecting physical things in workflow #1219

### LINKED ISSUE #1219: Performance issue in new project workflow
When running through the new project workflow as the dev (not admin) user - attempt to select physical things for your project.  There's a significant lag during selection.

jacobtylerwalls:
Avoid unnecessary GET requests when selecting physical things in the new project workflow. All the wanted information has already been fetched by the client.

@aarongundel I'd be grateful if you could let me know whether this solves the performance issue you experienced when running in Docker.

Closes #1219

### DIFF
--- a/afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
+++ b/afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
@@ -68,16 +68,16 @@ define([
         this.termOptions = [];
         this.value = ko.observableArray([]).extend({
             rateLimit: 100
-        });
-        this.startValue = ko.observableArray();
-        this.selectedResources = ko.observableArray([]);
-        this.addedValues = ko.observableArray();
-        this.removedValues = ko.observableArray();
+        });  // resourceinstanceid[]
+        this.startValue = ko.observableArray();  // Resource[]
+        this.selectedResources = ko.observableArray([]);  // Resource[]
+        this.addedValues = ko.observableArray();  // resourceinstanceid[]
+        this.removedValues = ko.observableArray();  // resourceinstanceid[]
 
         this.dirty = ko.pureComputed(function() {
             if (self.startValue() && self.value()){
-                return !!(self.startValue().find(x => !self.value().includes(x))
-                    || self.value().find(x => !self.startValue().includes(x)));
+                return !!(self.startValue().find(x => !self.value().includes(x.resourceinstanceid))
+                    || self.value().find(x => !self.startValue().map(v => v.resourceinstanceid).includes(x)));
             } else {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
```diff
@@ -89,15 +89,24 @@ define([
...
+                    let resource;
+                    if (self.dirty()) {
+                        resource = ko.unwrap(self.targetResources).find(
+                            r => r._id === ko.unwrap(action.value)
+                        )._source;
+                        resource = {
+                            ...resource,
+                            displayname: self.getStringValue(resource.displayname),
+                            displaydescription: self.getStringValue(resource.displaydescription),
+                        };
```
jacobtylerwalls, line 101: Decided to resolve the languages here, rather than in the templates.

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
```diff
@@ -121,11 +130,10 @@ define([
                         .filter(rr =>
                             rr.graph_id === "9519cb4f-b25b-11e9-8c7b-a4d18cec433a"
                         )
-                        .map(rr => rr.resourceinstanceid)
```
jacobtylerwalls, line 124: Instead of going to the server to get tile data, we just preserve it here. This required adjusting `self.startValue` to store resources instead of resourceinstanceids, but I kept `self.value` as an array of resourceinstanceids (better to compare on that, rather than on richer objects that might differ slightly).

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
```diff
@@ -68,16 +68,16 @@ define([
...
-        });
-        this.startValue = ko.observableArray();
-        this.selectedResources = ko.observableArray([]);
-        this.addedValues = ko.observableArray();
-        this.removedValues = ko.observableArray();
+        });  // resourceinstanceid[]
+        this.startValue = ko.observableArray();  // Resource[]
+        this.selectedResources = ko.observableArray([]);  // Resource[]
+        this.addedValues = ko.observableArray();  // resourceinstanceid[]
+        this.removedValues = ko.observableArray();  // resourceinstanceid[]
```
jacobtylerwalls, line 75: Then I clarified that here.

jacobtylerwalls (commented):
Should be working now, @aarongundel.

afs/media/js/views/components/workflows/create-project-workflow/add-things-step.js
aarongundel: This is interesting notation; makes me think of typescript.

aarongundel (approved):
This is fantastic @jacobtylerwalls - works great, major improvement within docker.

PR #1246, opened 2023-08-04
Fix SyntaxWarning in phys thing search workflow step

jacobtylerwalls:
Resolves error seen when running the project:
```py
/Users/jwalls/prj/afs/afs/views/physical_thing_search.py:23: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  has_filters = True if len(list(request.GET.items())) > 1 is not None else False
```

Luckily, `1 is not None` is equivalent to `1` in a comparison (`>`) context, so I don't think this was causing any issues.


njkim (approved):
Look good tome

PR #1248, opened 2023-08-05
Move temp_file view to the core arches, #9749

njkim:
This will move the temp_file view to the core arches.
Also, this will update `.gitignore` to ignore /archestemp/

This should be tested with https://github.com/archesproject/arches/pull/9854, and be merged after the core pr (`bulk_data_manager_improvements`).


chiatt:
Addressed in #1361

PR #1254, opened 2023-08-07
Fix 500 error from undefined manifest description #1228

### LINKED ISSUE #1228: Adding an existing manifest in upload manifest steps fails
During workflow steps in which users assign manifests to physical things (eg, the 'image' step in analysis areas or sample taking), creating and assigning a manifest works as expected. However, if a user opens the manifest manager and selects an existing manifest that is not served by Canteloupe but rather an external IIIF service, the step will not save a digital resource related to that manifest. Instead, the user will get the following error:

```
[03/Aug/2023 18:51:18] "POST /tile HTTP/1.1" 200 643
[03/Aug/2023 18:51:18] "GET /resource/descriptors/e492cc1c-76ca-4207-9101-0b34751b2337 HTTP/1.1" 200 357
2023-08-03 18:51:18,935 arches.app.views.tile ERROR    Unable to save. Please verify your input is valid [Tile id: ]                      [Exception message: TileValidationError: string error, {'en': {'direction': 'ltr'}}  - This is not a string. Unable to save.]                      [Exception trace: Traceback (most recent call last):
  File "/Users/cyrus/Documents/projects/GCI/arches4/arches/arches/app/views/tile.py", line 144, in post
    tile.save(request=request, transaction_id=transaction_id)
  File "/Users/cyrus/Documents/projects/GCI/arches4/arches/arches/app/models/tile.py", line 447, in save
    self.validate([], request=request)
  File "/Users/cyrus/Documents/projects/GCI/arches4/arches/arches/app/models/tile.py", line 340, in validate
    raise TileValidationError(_("{0}".format(error_instance["message"])))
arches.app.models.tile.TileValidationError: "string error, {'en': {'direction': 'ltr'}}  - This is not a string. Unable to save."
]
Internal Server Error: /tile
[03/Aug/2023 18:51:18] "POST /tile HTTP/1.1" 500 218
```

jacobtylerwalls:
Closes #1228

Keeps the analysis areas workflow in sync with the sample taking workflow, which guarded against the reported issue like this:
https://github.com/archesproject/arches-for-science-prj/blob/b14a0a60d30886c9e26d0b8be41d7a69b6aa5b46/afs/media/js/views/components/workflows/sample-taking-workflow/sample-taking-image-step.js#L95

***
Note: with the time remaining this afternoon, I'll look into factoring out a util that could be shared by the two files.

### DIFF
--- a/afs/media/js/views/components/workflows/analysis-areas-workflow/analysis-areas-image-step.js
+++ b/afs/media/js/views/components/workflows/analysis-areas-workflow/analysis-areas-image-step.js
@@ -7,8 +7,9 @@ define([
     'models/graph',
     'viewmodels/card',
     'templates/views/components/workflows/analysis-areas-workflow/analysis-areas-image-step.htm',
+    'views/components/workflows/stringUtils',
     'views/components/plugins/manifest-manager',
-], function(_, $, arches, ko, koMapping, GraphModel, CardViewModel, analysisAreasImageStepTemplate) {
+], function(_, $, arches, ko, koMapping, GraphModel, CardViewModel, analysisAreasImageStepTemplate, stringUtils) {
     function viewModel(params) {
         var self = this;
         params.pageVm.loading(true);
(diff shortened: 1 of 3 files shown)


chiatt (approved):
Looks good!

PR #1256, opened 2023-08-08
Add the relationship in upload xy data workflow, #1247

njkim:
Add relationship concept value to the related resource creation in upload xy data workflow, #1247


chiatt (approved):
Looks good!

PR #1258, opened 2023-08-09
Fix empty user in edit log

### LINKED ISSUE #1252: Delete button always raises PermissionDenied in Analysis Area workflow
1. Analysis area workflow
2. Complete some steps
3. Click Delete in top right and confirm

Result: always raises PermissionError even though it was the same user that created everything.

Traced it to EditLog entries being created without user id's.

### Traceback ###
```py
  File "/Users/jwalls/prj/arches/arches/app/utils/decorators.py", line 144, in wrapper
    raise PermissionDenied
django.core.exceptions.PermissionDenied
[07/Aug/2023 10:27:35] "POST /transaction/bb9fe741-6067-4a94-9764-a7bf6e813927/reverse HTTP/1.1" 403 135
```

jacobtylerwalls:
### Before ###
In the sample taking / analysis area workflows, it was possible to have a pending transaction where all of the EditLog entries lacked a user id. For non-superusers, this would prevent deleting the workflow, because there is a check that at least one EditLog entry exists for the transaction for the user attempting the deletion.

### Now ###
The user is included with the edit log entries.

Closes #1252

### DIFF
--- a/afs/tests.py
+++ b/afs/tests.py
@@ -1,0 +1,73 @@
+from django.core.management import call_command
+
+from django.test import TestCase
+from django.test.client import Client
+from django.urls import reverse
+
+from arches.app.utils.betterJSONSerializer import JSONSerializer
+from arches.app.models.models import Node, NodeGroup, ResourceInstance, ResourceXResource, TileModel
+from arches.app.models.tile import Tile
+
+def setUpModule():
+    call_command("packages", "-o load_package -s afs/pkg -y".split())
+
+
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
After discussing with @chiatt, adding the regression test and then futuring the work to get an action to run this on CI.

njkim (approved):
It looks great, works as expected.

njkim:
The test may need to be updated further to work generally. Otherwise the fix itself works as expected.

PR #1261, opened 2023-08-10
1259 Fix alert styling and input labels

### LINKED ISSUE #1259: Fix object selector label and alert
The project info step in the Upload XY step has 2 issues

First - the label needs to be changed to the following:

Select an Object from from your project
**Object must have at least one analysis area or sample location**

Second - the alert that shows up if a user picks an object without an analysis area should be aligned with the form
image

johnatawnclementawn:
Resolves #1259


afs/templates/views/components/workflows/select-phys-thing-step.htm
```diff
@@ -18,8 +18,11 @@
...
         <div class="widget-input-label">
             <span data-bind="if not: datasetRoute">
-                Select an Object
+                Select an Object from your project
             </span>
+            <div data-bind="if not: datasetRoute" style="font-style: italic; margin-bottom: 10px;">
+                Object must have at least one analysis area or sample location
+            </div>
             <span data-bind="if: datasetRoute == 'non-destructive'">
                 Select an Object that has an Area of interest
```
aarongundel: I'm encountering a scenario where I still see the old label and the new label.  This seems incorrect.  
Screenshot 2023-08-14 at 9 43 11 AM

aarongundel (changes requested):
Encountering some problematic behavior regarding the object selector labeling.

afs/templates/views/components/workflows/select-phys-thing-step.htm
johnatawnclementawn: Got some feedback from Cyrus, changes I just pushed should resolve this

afs/templates/views/components/workflows/select-phys-thing-step.htm
```diff
@@ -17,15 +17,22 @@
     <!--ko if: termFilter() -->
     <div style="padding: 5px; margin-top: 15px;">
         <div class="widget-input-label">
-            <span data-bind="if not: datasetRoute">
-                Select an Object
+            <!-- ko if: datasetRoute == 'non-destructive' -->
+            <span>
```
njkim: The combination of span and div in parallel does not seem to be necessary and less readable. Probably we can two div's.

afs/templates/views/components/workflows/select-phys-thing-step.htm
```diff
@@ -17,15 +17,22 @@
     <!--ko if: termFilter() -->
     <div style="padding: 5px; margin-top: 15px;">
         <div class="widget-input-label">
```
njkim: The label doesn't appear in the other workflows that use this step, eg. sample taking wf, analysis area wf.

afs/templates/views/components/workflows/select-phys-thing-step.htm
```diff
@@ -17,15 +17,22 @@
...
+            <div style="font-style: italic; margin-bottom: 10px;">
+                Object must have at least one analysis area or sample location
+            </div>
+            <!-- /ko -->
+            <!-- ko if:  datasetRoute == 'destructive' -->
+            <span>
                 Select a Sample of interest
             </span>
+            <div style="font-style: italic; margin-bottom: 10px;">
+                Object must be a sample
```
njkim: Maybe we don't need this. The label says select a sample :)

njkim (changes requested):

afs/templates/views/components/workflows/select-phys-thing-step.htm
johnatawnclementawn: Hey @njkim, what should those labels say in the other wf's? The current labels don't show up because the `datasetRoute` is not initialized for them.

njkim: Before your change, I believe the wf that doesn't have a route was handled with a generic label.

aarongundel (approved):
LGTM

njkim (approved):
It looks good.

PR #1262, opened 2023-08-10
Update type label in new project workflow #1234

### LINKED ISSUE #1234: Allow Project to Save without Activity Type in New Project Workflow
First step in new project workflow fails if user does not select a Project Activity Type.  This is not a required node, so the project should save without error if the user does not add a type.

jacobtylerwalls:
**Before**
Before, in the New Project Workflow, the project type was mislabeled, see #1234.

**Now**
Screenshot 2023-08-10 at 2 01 10 PM

Closes #1234

### DIFF
--- a/afs/media/js/views/components/workflows/create-project-workflow/project-name-step.js
+++ b/afs/media/js/views/components/workflows/create-project-workflow/project-name-step.js
@@ -99,6 +99,12 @@ define([
                 return;
             }
     
+            if (!self.typeValue()) {
+                params.form.error(new Error("Missing Required Value"));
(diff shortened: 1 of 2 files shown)


afs/media/js/views/components/workflows/create-project-workflow/project-name-step.js
```diff
@@ -99,6 +99,12 @@ define([
                 return;
             }
     
+            if (!self.typeValue()) {
```
chiatt, line 102: I see this follows the same pattern as the `Project Name` node, but we should actually handle for missing required values using the response from the server.  That way the error message will be properly localized. We can address this for both nodes in another PR. This fix prevents the step from breaking, though, so it's a good stop-gap.

chiatt (approved):
Looks good to me!

PR #1264, opened 2023-08-14
Fix the report workflow annotation step, #1223 

### LINKED ISSUE #1223: Report workflow fails if project does not have any annotations
It seems like nearly all reports would have annotations, but we should probably handle for their absence.

image

njkim:
Make some Improvements on the project report workflow add-annotation step.
1. Fix the issue and also improve the ui with the case the annotation is not available, re #1223
2. Fix the issue with annotation isn't properly updated
3. Improve the step to handle multiple canvases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ Procfile
 afs/settings_docker.py
 CACHE
 afs/export_deliverables/
+afs/archestemp/
 afs/staticfiles
 afs/webpack/webpack-stats.json
 node_modules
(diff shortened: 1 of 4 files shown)


chiatt (changes requested):
When I load a manifest with two canvases, the canvases load and switch as expected, but the thumbnails are identical. 

image

njkim:
Now it can handle the multiple canvases from one or more manifest(s).

> When I load a manifest with two canvases, the canvases load and switch as expected, but the thumbnails are identical.

chiatt (approved):

Issue #1265, opened 2023-08-15
Sample Taking WF - Sample Info step hangs after saving edits to existing sample


johnatawnclementawn:
1. Sample taking workflow
2. Sample info step: draw a feature, save
3. Sample info step loads, edit feature that was just created & save
4. Loading screen will hang on `Saved.` - requires a refresh to return to Sample info step

jacobtylerwalls:
Hey @njkim is this ready to close?

PR #1267, opened 2023-08-17
Fix the sampling annotation save error, #1265

njkim:
Fix the issue with accessing canvas observable, #1265


njkim:
Closing, #1270 overwrites this

PR #1268, opened 2023-08-17
Add compatibility with Arches installed apps changes in 7.5

jacobtylerwalls:
Refs archesproject/arches#9426

Next step: #1275 makes changes needed for actually distributing this application to other consumers.

~Discussion point: should we also remove manage.py from this repo to clarify it's never intended to be run on its own? Especially if the app no longer runs normally after moving the package data as described above.~
Let's keep it so that we can `manage.py test`.


afs/package.json
```diff
@@ -10,19 +10,29 @@
         "@uppy/dashboard": "^2.4.0",
         "@uppy/drag-drop": "^2.1.2",
         "@uppy/progress-bar": "^2.1.1",
-        "arches": "archesproject/arches#dev/7.4.x",
+        "arches": "archesproject/arches#9304-cbyrd-arches-installed-packages",
```
jacobtylerwalls: TODO: adjust this as soon as the branch is merged in

chiatt (changes requested):
This looks great, but It looks like we can safely delete this file `afs/webpack/webpack-node-modules-aliases.js`.  Could you remove it?

chiatt (approved):

chiatt:
It appears `afs/webpack/webpack-node-modules-aliases.js` was added to the project by the `update_project` [command](https://github.com/archesproject/arches/blob/deadd9d812788e767d45e2806adab9a42c042d38/arches/management/commands/updateproject.py#L28).  Given that, we should probably leave it in AFS until we're sure it's not needed.

PR #1271, opened 2023-08-18
The dropdown to add the object of interest for report, #1266

### LINKED ISSUE #1266: Allow user to select objects of study in reporting workflow
In the reporting workflow, a user should be able to identify the object(s) of study. Elements of the report will be limited to these physical things and their related resource data. It will also allow the report template developer to identify which object(s) should be the focus of the report.

### LINKED ISSUE #1310: Multiselect resource instance widget shows question mark instead of resource icon in first step of project reporting workflow

njkim:


afs/templates/views/components/workflows/project-report-workflow/project-report-select.htm
```diff
@@ -12,4 +12,25 @@
...
     }"></div>
+    <!-- ko if: hasSetWithPhysicalThing() === false -->
+    <div class="tabbed-workflow-step-information-box widget-input-label" style="text-align: center; margin-top: 16px; height:fit-content; background-color: #fde9ad;">
+        This project has no associated Objects. Your project must have an associated Object Collection with an Object before proceeding.
+    </div>
+    <!-- /ko -->
+    <!-- ko if: termFilter() || physicalThingsValue() -->
+    <div style="padding: 5px; margin-top: 15px;">
+        <div class="widget-input-label">
+            <span>{% trans "Select an Object" %}</span>
```
chiatt: Can you change to `Select Object(s) to include in your report`

chiatt (changes requested):
This is looking really good. Could you just update the label on the object dropdown to indicate that they can select more than one thing? e.g. "Select Object(s) to include in your report"
Also, could you remove the question marks in the selected physical things? 
Screenshot 2023-08-30 at 3 00 47 PM

afs/templates/views/components/workflows/project-report-workflow/project-report-select.htm
njkim: The title has updated.
But the question mark is a standard behavior of the resource-instance widget when widget can't get the iconclass for some unknown reason (possibly because the list is from term filter)

chiatt (approved):

jacobtylerwalls:
Hey @njkim there are nontrivial merge conflicts with dev/1.1.x, would you be able to take a look and update? Thanks!

njkim:
> Hey @njkim there are nontrivial merge conflicts with dev/1.1.x, would you be able to take a look and update? Thanks!

@jacobtylerwalls This should be good now.

arches_for_science/templates/views/components/workflows/project-report-workflow/project-report-select.htm
```diff
@@ -12,4 +12,25 @@
             disabled: locked,
         }
     }"></div>
+    <!-- ko if: hasSetWithPhysicalThing() === false -->
+    <div class="tabbed-workflow-step-information-box widget-input-label" style="text-align: center; margin-top: 16px; height:fit-content; background-color: #fde9ad;">
+        This project has no associated Objects. Your project must have an associated Object Collection with an Object before proceeding.
+    </div>
+    <!-- /ko -->
+    <!-- ko if: termFilter() || physicalThingsValue() -->
+    <div style="padding: 5px; margin-top: 15px;">
```
jacobtylerwalls: The 5px padding ends up indenting it, which might be intentional, but then the right end sticks out further. Do you think it might look better without the padding?

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-project-files.js
```diff
@@ -10,10 +10,16 @@ define([
...
 
         this.projectValue = params.projectId;
+        const physicalThings = params.physicalThings;
 
         const observationGraphId = '615b11ee-c457-11e9-910c-a4d18cec433a';
         const digitalResourcegGraphId = '707cbd78-ca7a-11e9-990b-a4d18cec433a';
+        const collectionGraphId = '1b210ef3-b25c-11e9-a037-a4d18cec433a';
+        const physicalThingGraphId = '9519cb4f-b25b-11e9-8c7b-a4d18cec433a';
         const fileNodeId = '7c486328-d380-11e9-b88e-a4d18cec433a';
+        const objectObeservedNodeId = "cd412ac5-c457-11e9-9644-a4d18cec433a";
```
jacobtylerwalls: Perhaps fix the spelling here

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
```diff
@@ -37,6 +46,24 @@ define([
...
 
         const generateReport = async(template) => {
+            const relatedObjects = await getRelatedResources(projectId);
+            const collections = relatedObjects.related_resources.filter(rr => rr.graph_id == collectionGraphId);
+            const allPhysicalThingsResponse = collections.map(async(collection) => {
+                const collectionRelatedResources = await getRelatedResources(collection.resourceinstanceid);
+                return collectionRelatedResources?.related_resources.filter(rr => rr.graph_id == physicalThingGraphId);
+            });
+
+            self.physicalThings = [].concat(...(await Promise.all(allPhysicalThingsResponse))).filter(res => {
```
jacobtylerwalls, line 56: Tried a couple projects, and generating the reports all failed like this on the last step. Let me know if you see the same thing.
```py
2023-09-21 15:07:13 django.core.exceptions.ValidationError: ['“” is not a valid UUID.']
2023-09-21 15:07:13 [21/Sep/2023 19:07:13] "GET /api/bulk_disambiguated_resource_instance?v=beta&resource_ids= HTTP/1.1" 500 190147
```

jacobtylerwalls (changes requested):
I pushed a commit to fix #1310, so suggest pulling before doing more work. One issue prevented me from testing, will look forward to taking another look once it's fixed!

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-project-files.js
njkim: thanks, good catch

arches_for_science/templates/views/components/workflows/project-report-workflow/project-report-select.htm
njkim: The padding removed. it seems its parent div has a padding which I thought as a sibling. Thanks.

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
njkim: @jacobtylerwalls I don't see this happening... hmmm. not sure what causes the issue.
Maybe you have something that I don't or vice versa.

jacobtylerwalls: Aha! I was able to reproduce the same problem on dev/1.1.x, so it's not on this branch. 👍 

My issue was I didn't have any observations. How should I create some?

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-project-files.js
```diff
@@ -40,26 +46,39 @@ define([
...
-                .then(response => response.json())
-                .then(json => {
-                    self.projectName = json.resource_instance.displayname;
-                    projectObservations = json.related_resources.filter(res => res.graph_id == observationGraphId);
+            const projectResponse = await window.fetch(arches.urls.related_resources + self.projectValue  + "?paginate=false");
+            const projectJson = await projectResponse.json();
+
+            const collectionForProject = projectJson.related_resources.find(res => res.graph_id === collectionGraphId).resourceinstanceid;
+            const collectionResponse = await window.fetch(arches.urls.related_resources + collectionForProject  + "?paginate=false");
+            const collectionJson = await collectionResponse.json();
```
jacobtylerwalls, line 55: Would it be possible to process these two requests in parallel, or do they depend on each other?

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
jacobtylerwalls, line 56: a cosmetic suggestion at your option
```suggestion
            self.physicalThings = [...(await Promise.all(allPhysicalThingsResponse))].filter(res => {
```

arches_for_science/media/js/views/components/workflows/project-report-workflow/project-report-select.js
```diff
@@ -3,30 +3,93 @@ define([
...
+        this.projectValue.subscribe(function(val){
+            if (!initializing) {
+                self.physicalThingsValue(null);
+            }
+            initializing = false;
+            if (val) {
+                var res = resourceUtils.lookupResourceInstanceData(val);
+                res.then(
+                    function(data){
+                        self.projectNameValue(data._source.displayname);
```
jacobtylerwalls, line 44: Wondering if this is normally how we deal with string data. Or, should we get it from the tile to so we can present it in the user's language? Maybe we're not worrying about that for AFS.

arches_for_science/media/js/views/components/workflows/project-report-workflow/project-report-select.js
```diff
@@ -3,30 +3,93 @@ define([
...
+                        let setTileResourceInstanceIds;
+                        let setTile = data._source.tiles.find(function(tile){
+                            return tile.nodegroup_id === self.physThingSetNodegroupId;
+                        });
+                        if (setTile && Object.keys(setTile.data).includes(self.physThingSetNodegroupId) && setTile.data[self.physThingSetNodegroupId].length) {
```
jacobtylerwalls: cosmetic suggestion if you like
```suggestion
                        if (setTile?.data?.[self.physThingSetNodegroupId]?.length) {
```

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
```diff
@@ -69,7 +95,8 @@ define([
                 analysisAreas,
                 observationDetails: [...Object.values(observationDetails)],
                 projectDetails: [...Object.values(projectDetails)],
-                physicalThingsDetails: physicalThingsDetailsArray
+                physicalThingsDetails: physicalThingsDetailsArray,
+                objectOfStudyDetails: objectOfStudyDetailsArray
```
jacobtylerwalls, line 99: Should the child objects (e.g. samples) appear in the report if I've selected the parent? When I select "Jefferson Memorial" as the object of interest, I also get its samples in the report:
Introduction

jacobtylerwalls (approved):
Tested, everything works well. Just leaving a few questions for you to consider. 👍

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-project-files.js
njkim: The user selects the project and the objects of interest. The collection for project needs to be searched using the project selected. So, I believe that those two requests have to be run in sequence.

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
njkim: I believe thus is intentional since the sample is a real object unlike the areas.
But the requirements are still being refined, so it may be changed in the future, as well as the above (the request on collection) For example, do all the physical things to be included or just the one that the user selected...)
This should be good for now.

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
njkim: I believe thus is intentional since the sample is a real object unlike the areas.
But the requirements are still being refined, so it may be changed in the future, as well as the above (the request on collection) For example, do all the physical things to be included or just the one that the user selected...)
This should be good for now.

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
njkim: The Promise.all returns an array of arrays.
If the response spreads in [], we will end up having an array of arrays again (The above 2nd cod block).
If the response spreads and concats then we will get an array of the objects. (I think that's what we want)

arches_for_science/media/js/views/components/workflows/project-report-workflow/project-report-select.js
njkim: Good question. I would think that the resourceUtil should return properly localized string of the displayname. But it is quite possible that it may not. (It looks like it's been used in various places in afs for a while).
Like you imagine, it has not been a concern for AFS as a project but maybe something to think about for the future (as an app).

arches_for_science/media/js/views/components/workflows/project-report-workflow/download-report.js
jacobtylerwalls: ~I swapped in the suggestion and it still returns an array of objects. I think it's because Promise.all() returns an array of objects/resolved promises.~

Fine to leave this as is, I just tend to ask questions like this when I'm not sure I have the right understanding of what's going on :-)

jacobtylerwalls: ~Oh! The filter is chained to the concat, sorry. Now I see what it's doing.~

jacobtylerwalls: Thanks for your patience while I worked through this pattern; I'd never seen this strategy for reducing dimensionality before through concat!

jacobtylerwalls:
Looks like the CI failure is https://github.com/python-openxml/python-docx/issues/1256, so I'll open a PR to pin the package with the issue.

jacobtylerwalls (approved):

Issue #1273, opened 2023-08-21
Adding physical things to project does not set userid on edit log entries


jacobtylerwalls:
Similar to #1252, the lack of a user in edit log entries can prevent workflow deletion in phys thing workflows.

### To reproduce ###
0. Begin with a project that has a physical thing already
1. Login as dev/dev
2. Manage objects workflow
3. Remove the object, advance to final screen
4. Delete workflow

### Result ###
```
afs7             | [21/Aug/2023 21:34:23] "GET /resources/6feff107-af72-4c3c-aa34-5c1aeb97eb8b?format=json&compact=false&v=beta HTTP/1.1" 200 1132
afs7             | Forbidden (Permission denied): /transaction/2f8ff0b4-2849-4ee6-af40-cf0f39abdeda/reverse
afs7             | Traceback (most recent call last):
afs7             |   File "/root/.local/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
afs7             |     response = get_response(request)
afs7             |   File "/root/.local/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
afs7             |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
afs7             |   File "/root/.local/lib/python3.10/site-packages/django/views/generic/base.py", line 70, in view
afs7             |     return self.dispatch(request, *args, **kwargs)
afs7             |   File "/root/.local/lib/python3.10/site-packages/django/utils/decorators.py", line 43, in _wrapper
afs7             |     return bound_method(*args, **kwargs)
afs7             |   File "/web_root/arches/arches/app/utils/decorators.py", line 143, in wrapper
afs7             |     raise PermissionDenied
afs7             | django.core.exceptions.PermissionDenied
```

**Reason**
<img width="457" alt="Screenshot 2023-08-21 at 5 36 57 PM" src="https://github.com/archesproject/arches-for-science-prj/assets/38668450/98bc4a07-49fc-4083-aaa0-8a9384f364e7">

PR #1280, opened 2023-08-29
Fix typo, re #1273

chiatt:
Fixes minor typo in save call, re #1273


njkim (approved):
Looks good. good catch!

PR #1275, opened 2023-08-23
Modularize into an installable app

jacobtylerwalls:
Make changes needed for installing this project as an "arches installed application" in 7.5.

Depends on #1268

PR #1283, opened 2023-08-31
Small update for the matching arches version

njkim:


jacobtylerwalls:
@njkim I think this is handled in #1268 and #1275, just FYI

njkim:
> @njkim I think this is handled in #1268 and #1275, just FYI
Oops :)
Thanks, I missed that.

PR #1285, opened 2023-09-01
Correct concept values for primary name/title

### LINKED ISSUE #1284: Sampling info and sample phys thing tiles use incorrect concepts for name
The load fails the the following error: 
```
concept error, preferred terms - This is an invalid concept prefLabel, or an incomplete UUID. Unable to save. | preferred terms
```

This error does not occur if data is loaded using branch excel.
Removing that concept label from the worksheet allows the excel file to load without error.

image

To reproduce - 
1. Create some physical things using workflows
2. Export them through search export in tile excel format
3. Delete your physical things from your project
4. Attempt to reload them using tile excel import

image

jacobtylerwalls:
Fixes #1284

### Testing instructions ###
- [ ] Check the graph editor for Sample Info, take note of the concept values allowed for name
- [ ] Check this PR sets them (primary title, alternate titles)
- [ ] Check the graph editor for Physical Thing, take note of the concept values allowed for name
- [ ] Check this PR sets a reasonable choice (primary name)

### DIFF
--- a/arches_for_science/media/js/views/components/workflows/sample-taking-workflow/sampling-info-step.js
+++ b/arches_for_science/media/js/views/components/workflows/sample-taking-workflow/sampling-info-step.js
@@ -152,7 +152,7 @@ define([
 
         this.saveSamplingName = function() {
             const languageValueId = ['bc35776b-996f-4fc1-bd25-9f6432c1f349']; //English
-            const prefLabelIds = ["7d069762-bd96-44b8-afc8-4761389105c5","8f40c740-3c02-4839-b1a4-f1460823a9fe"]; //[primary title, preferred terms];
+            const prefLabelIds = ["7d069762-bd96-44b8-afc8-4761389105c5", "0798bf2c-ab07-43d7-81f4-f1e2d20251a1"]; // [primary title, alternate titles]
             const samplingNameTileData = {
                 "tileid": ko.unwrap(self.samplingNameTile) || "",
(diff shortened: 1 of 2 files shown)


chiatt:
We may need a migration to modify any existing records that users have already created. I don't think we necessarily need to do this as part of this PR, however the SQL to make that change would be something like this:

```
UPDATE tiles
SET tiledata = jsonb_set(tiledata, '{b9c1d7ab-b497-11e9-9ab7-a4d18cec433a, 0}', '"0798bf2c-ab07-43d7-81f4-f1e2d20251a1"'::jsonb)
WHERE EXISTS (
  SELECT 1
  FROM jsonb_array_elements(tiledata->'b9c1d7ab-b497-11e9-9ab7-a4d18cec433a') AS element
  WHERE element::text = '"8f40c740-3c02-4839-b1a4-f1460823a9fe"'
);
```

chiatt:
Thanks for your Testing instructions @jacobtylerwalls. When going through them I noticed that the default value for the name type node is also `preferred terms` so we may need to also make a minor change to the graph design. I'm going to confirm this before we make any more changes.

chiatt:
It turns out that we need to continue to use 'preferred terms'. The reason why this caused issues with the bulk loader is because `preferred terms` concept was assigned as the default value despite the fact that 'preferred terms' was not included in the collection (re #1330).

PR #1286, opened 2023-09-02
Rename to `arches-for-science`

### LINKED ISSUE #1274: Rename package from "afs"
Before asking users to install this project as an arches installed app, we'll need a better name than "afs", which already exists on PyPI.

jacobtylerwalls:
Fixes #1274

#### Didn't change ####
ALLOWED_HOSTS = ["afs.local", "10.0.2.2", "localhost"]
afs7

#### didn’t change because should be allowed to be arbitrary anyway ####
ENV APP_ROOT=${WEB_ROOT}/afs


jacobtylerwalls:
`arches_for_science` will be the name in python and postgres, but in pip/pypi it will be `arches-for-science`.

.gitignore
```diff
@@ -1,26 +1,27 @@
 *.pyc
-afs/media/node_modules
```
aarongundel, line 2: As part of this PR, it might be worth it to just ignore anything in the "afs" directory.  That way other incoming PRs don't get things twisted and we end up with arches_for_science and afs

jacobtylerwalls: I like it

jacobtylerwalls: [658d826](https://github.com/archesproject/arches-for-science-prj/pull/1286/commits/658d826322cb882bff0a5ca2b30bcf582f22e8ec)

chiatt (approved):
Looks good to me!

PR #1287, opened 2023-09-05
xy data loader

### LINKED ISSUE #1218: Two-Dimensional (Spectral) Digital Data Loader
Arches for Science implements a workflow for uploading digital files obtained from laboratory instruments. Certain instruments create files that typically include header information and a long series of x/y data that represent the results of an instruments measurements. GCI conservation scientists may export such “two-dimensional” datasets as ASCII files and import them into Arches as part of the upload dataset workflow.

The GCI’s conservation laboratory includes a number of instruments that create two-dimensional data, each with its own unique header information and representation of the x- and y-value data. Currently, AfS requires a specific dataset loader to process each instrument’s uniquely formatted two-dimensional data.

GCI conservation scientists have requested that AfS include a “generic” ASCII data file loader that would allow them to configure how Arches should import and process the information in the file. As part of this task, Farallon will develop a data loader that allows users to preview a file, indicate where in the file x/y values should begin to be imported, and the type of mathematical transformation (if any) to apply to the x or y data value as part of the import process.

### LINKED ISSUE #1318: Upload step on Upload XY Data provides no feedback on failure
If the file upload on the Upload XY Data Workflow fails, the step provides no feedback to indicate that's the case.  At minimum, we should probably tell the user that the file failed to upload.

aarongundel:
Adds features relevant for #1218 .  Allows for dynamically formatted files to be read.  Deprecates other afs formatters (raman reader, xrf reader, fors reader) which are retained for backward compatibility.  Adds the concept of an "importer" - an attribute stored on the file tile that points to a database record that specifies a configuration that gets passed to the new xy reader.  That configuration tells the XY reader how to parse the file.  The reader itself uses the XY Parser utility to do the dirty work of parsing the file.  Supports the concept of transformations - provided multiple Y values, transforming those Y values via known methods and functions.  As part of this PR, the "mean" or "average" transformation is included (as well as a "basic" transformation which selects the first y value if multiples are given.

Resolves #1218 

I suspect we will need to include some kind of import functionality in Arches for Science to transfer these importers between installations and perhaps provide a well-known set, but that is outside the scope of this PR.


aarongundel:
This PR can be squashed.

aarongundel:
This PR requires  https://github.com/archesproject/arches/pull/10003

aarongundel:
The previously mentioned PR needs to merged into dev/7.5.x for this to work.

arches_for_science/views/workflows/upload_dataset/select_dataset_files_step.py
```diff
@@ -139,11 +138,10 @@ def post(self, request):
                         file_data["renderer"] = next(
                             (renderer["id"] for renderer in settings.RENDERERS if renderer["name"] == "pdfreader"), None
                         )
-                    elif dataset_default_format is not None:  # instrument was given by zip file name
+                    elif file_data["name"].split(".")[-1] == "dx" or file_data["name"].split(".")[-1] == 'txt':  # instrument was given by zip file name
```
chiatt, line 141: It seems like afs should have a setting with a list of possible 'txt' extensions.

arches_for_science/media/js/utils/xy-parser.js
```diff
@@ -0,0 +1,54 @@
...
+                    values = lines.slice(config?.headerFixedLines);
+                } else {
+                    values = workingText.trim().split('\n'); 
+                }
+            } catch(e) {
+                values = workingText.trim().split('\n');
+            }
+            const delimiterCharacter = config?.delimiterCharacter ?? ',';
+
+            const valueRegex = (delimiterCharacter.length < 2) ? new RegExp(`[${delimiterCharacter}\\s]+`) : new RegExp(`${delimiterCharacter}`);
```
chiatt, line 42: This regex silently failed when I used `**` as a delimiter. This is a pretty unlikely delimiter, but it would be nice to have a message on the front end stating that the delimiter cannot be parsed - just in case.

chiatt (approved):
This looks great. Added just a couple of comments, but I think I'll write them up as other issues.

PR #1298, opened 2023-09-13
Install arches from pypi rather than github

chiatt:
Install arches from pypi rather than github repo


jacobtylerwalls (approved):
tested with `pip install .`, and it pulled 7.5.0a0 from PyPI 👍

PR #1302, opened 2023-09-15
Relax minimum arches version

jacobtylerwalls:
Refs archesproject/arches@cb212a6

```py
arches.app.utils.compatibility.CompatibilityError: 'This project is incompatible with Arches 7.5.0a0.'
```


chiatt (approved):
Looks good. We should probably look at skipping this check in core arches if compatibility is enforced by Python (as it would be for any arches application).

PR #1307, opened 2023-09-15
Unselect sample on save #1249

### LINKED ISSUE #1249: Sample shape disappears if edited after initially created
1. Sample Taking Workflow
2. Sample info step: draw a line, save
3. Edit the points of the same line
Result: line "vanishes". Refreshing the page will restore it.

jacobtylerwalls:
In addition to fixing the underlying bug where a sample would be cleared from the canvas once reselected, this also brings the sample workflow into agreement with the analysis workflow. Now, in both workflows, on save, nothing is automatically selected.

Suggest retesting the workflow from #1249. Ensure a sample can be re-edited after saving, and that all changes persist through completion of the workflow.

Closes #1249


johnatawnclementawn (commented):
LGTM, there are further instabilities in this workflow that are described in a different ticket

johnatawnclementawn (approved):
LGTM, there are further instabilities in this workflow that are described in a different ticket: #1321

PR #1315, opened 2023-09-18
removes cruft after appification #1299

### LINKED ISSUE #1299: Clean up from Application Modification
AfS needs some cleanup after the application modification.  

Of the major changes for this issue, we need to

* remove all docker related files
* remove index page
* remove images related to index
* look for anything else that no longer useful.  

Do not remove settings.py - that should be imported in the project's settings.py

chrabyrd:
sister ticket to: https://github.com/archesproject/disco/pull/15


chrabyrd:
I may have been a bit heavy-handed in the removals/changes, but the application still worked for me with this package. Please reach out if you notice any destabilization!

arches_for_science/test_settings.py
```diff
@@ -3,7 +3,6 @@
 # Depends on /arches being on the python path, as arches.tests is not importable
 from tests.test_settings import *
 
-from arches_for_science.settings import INSTALLED_APPS, ROOT_URLCONF, APP_NAME
```
jacobtylerwalls, line 6: This was added recently just for the sake of the project unit tests, so with luck that's all we'll need to get them passing again 🤞

arches_for_science/settings.py
```diff
@@ -1,348 +0,0 @@
...
-    #     "component": "views/components/cards/file-renderers/colladareader",
-    #     "ext": "dae",
-    #     "type": "",
-    #    "exclude": [],
-    # },
-]
-
-X_FRAME_OPTIONS = "SAMEORIGIN"
-
-FORMATS = [
```
jacobtylerwalls, line 303: I very much like the idea of shipping a smaller settings.py, but removing the whole thing undoes archesproject/disco@95a5b8aef03b931f244e65549e408462257ece11. Maybe you have a recommendation for a very minimal set of settings we could keep?

chrabyrd: if possible we should discuss why an arches app would need a settings file? Django apps usually don't and instead require the project to adjust its settings file. And if it's only there to keep testing happy I think it would be best to investigate a way around that.

jacobtylerwalls: I'm open to different ideas here.

In terms of whether it's common to package settings.py, I can't speak to what's more common, but two Django apps I've used regularly and that we use at Farallon both package their settings.py (well, one, [django-oauth-toolkit](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/settings.py), explicitly says they borrowed the approach from the other, [django-rest-framework](https://github.com/encode/django-rest-framework/blob/master/rest_framework/settings.py)). [Django-guardian](https://github.com/django-guardian/django-guardian/blob/77de2033951c2e6b8fba2ac6258defdd23902bbf/guardian/conf/settings.py#L4) also does. I don't have a strong opinion about whether afs should also, I'm just saying it doesn't strike me as an anti-pattern right off the bat.

So there's two parts here, one is just getting the test suite to recognize the URLs config, and the other is shipping python constants that will be useful to consumers of arches for science, some of whom may not describe themselves as comfortable configuring details in Django (this is just an early impression from my very short tenure here so far!)

For the tests, we're already using a test_settings.py, so we can just define the few settings that we need directly there. I wasn't sure if you were suggesting something else. A good cleanup would be to exclude test_settings.py from the package manifest.

For the python constants like `FORMATS`, I guess it's a judgment call whether we should expect users to copy/paste it from a bootstrapping document or release note, versus whether we should advise people in that very note to import it from arches_for_science.settings (or arches_for_science.constants, if you prefer)? I don't know the `FORMATS` setting very well. @aarongundel might have thoughts.

chrabyrd: I hear what you're saying. It's not an anti-pattern per se, but I think the way we're including it arches apps should be refactored. The settings file included in your examples are wholly contained to the package, I think that's not entirely possible with arches apps.

What I've noticed from DISCO is that in the project's `settings.py` file, we're no longer importing settings from arches core. That has been moved to the arches-for-science app. To me this is an anti-pattern. Also having multiple arches apps importing their own `settings.py` files into a project could result in some confusing situations around static files, templates, etc.

I'll create a sister ticket in DISCO and update the project's settings file, along with updating `test_settings.py` to function with the changes and `README.md` to log the changes that need to be made a project's settings file.

jacobtylerwalls:
> we're no longer importing settings from arches core. That has been moved to the arches-for-science app. To me this is an anti-pattern.

+1

> Also having multiple arches apps importing their own settings.py files into a project could result in some confusing situations around static files, templates, etc.

absolutely. Looking forward to your changes here!

jacobtylerwalls (approved):
Thanks for this, these readme changes are great.

PR #1324, opened 2023-09-21
Don't show broken report generation as 'normal' - #1323

### LINKED ISSUE #1323: Project Report Workflow downloads can be broken
Project Report Workflow should not show "Download" button if the request to genereate the report fails.  It should inform the user of the failure.

aarongundel:
fixes #1323


chiatt (approved):
Looks good to me!

Issue #1334, opened 2023-10-03
There are invalid entries in the thesauri / the collections.


njkim:
There are invalid entries in concepts/collection xmls that prevent the package loading fail with core 7.5.x and afs 1.1.x.

chiatt:
Addressed with #1330

PR #1335, opened 2023-10-03
Fix the invalid entries in the rdm collections, #1334

njkim:
Fix the issue with invalid entries in the rdm collection which prevents the package to reloading due to the recent improvement in the core arches (https://github.com/archesproject/arches/pull/10100).


chiatt (approved):
Looks good to me!

PR #1336, opened 2023-10-03
fix for chemical analysis workflow

### LINKED ISSUE #1326: Chemical Analysis Workflow broken in disco
The CA workflow is broken in disco.  Looks like the reason is that the companion modification for uppy made in afs's webpack.common.js didn't make it into disco.  However, there the PR for companion's fix did merge into uppy - so it may be as simple as upgrading uppy to fix this.  If not, we'll have to modify the webpack process to replace companion again.

aarongundel:
fixes various issues with the chemical analysis workflow/uppy implementation.

* fixes inherent race condition that required files to be fairly large to beat the tile creation process.  
* fixes the lack of respect for django storages AWS_S3_FILE_OVERWRITE option 
* changes creation process.  

On dragging a new file in to be uploaded, that file will _first_ be created in arches and saved to s3 - as a zero byte file.  Then, that updated path (if need be due to AWS_S3_FILE_OVERWRITE) will be used in uploading the file.


arches_for_science/views/s3.py
```diff
@@ -89,12 +112,43 @@ def batch_sign(request, uploadid):
...
+        file_name = request.GET.get("filename")
+
+        if(file_name.startswith(KEY_BASE)):
+            key = file_name
+        else:
+            key = KEY_BASE + "/" + file_name
+
+        fields={}
+        #fields = dict(('x-amz-meta-{}'.format(key[9:len(key)-1]), value) for (key, value) in all_items if key.startswith('metadata') and key != 'metadata[type]')
+        #fields['content-type'] = content_type
```
chiatt: Looks like this can be removed

arches_for_science/views/s3.py
```diff
@@ -89,12 +112,43 @@ def batch_sign(request, uploadid):
                 )
             )
         return JsonResponse(urls, safe=False)
-    else:
-        return HttpResponseNotAllowed()
 
+@method_decorator(can_edit_resource_instance, name="dispatch")
+class S3UploadView(BaseManagerView):
+    """"""
```
chiatt: Empty docstring

chiatt (changes requested):
This works great - just found a little cruft that can be removed.

chiatt (approved):

PR #1340, opened 2023-10-09
Fix the logic to save text (prj name), #1332

### LINKED ISSUE #1332: Undoing a change to Project Name filed shows `[Object object]`
1. New project workflow
2. Enter a name
3. Click undo before saving

Screenshot 2023-09-11 at 4 00 15 PM

Probably sources to core arches.
Found while testing archesproject/arches#10096

njkim:
Fix the issue of showing [object Object] when the user undo the edit to initial (empty string), #1332


chiatt (approved):
Looks good!

Issue #1330, opened 2023-09-29
Add preferred terms to Name Type - Physical Thing collection


chiatt:
The preferred terms concept needs to be added to the 'Name Type - Physical Thing' collection in the RDM. We also need a migration that will add the concept to the collections of existing afs implementations.

chiatt:
@njkim I spoke with @annabelleee and @Habennin today and they said we should move the preferred terms concept under Name Types - Generic, then we can add it to any name type collection as needed.
1. Remove the `preferred terms` concept from under `Personal Name Types`
2. Add `preferred terms` under `Name Types - Generic`
3. Add `preferred terms` to any collection where it is missing
4. Be sure any graphs using `preferred terms` as a default value are updated if the conceptid has changed
5. Export the graphs and reference data so that the afs package is updated

PR #1341, opened 2023-10-10
The collection name type, #1330

njkim:
Add collection name type when creating a new project collection, #1330


chiatt:
Apologies @njkim - I should have provided some detail in the ticket. 'Collection' in this case refers to the concept collection rather than the collection resource model. We also need a migration that will add the concept to the collections of existing afs implementations.

njkim:
closed. the rdm collections will be updated, instead.

Issue #1329, opened 2023-09-29
Add setting with list of txt extensions


chiatt:
Setting should be applied here, replacing hard-coded .dx and .txt extensions.

https://github.com/archesproject/arches-for-science/pull/1287#discussion_r1340597611

PR #1348, opened 2023-10-12
The instrument text format setting, #1329

njkim:
Add settings for instrument file format the projects is using, #1329


chiatt (approved):
Looks good!

PR #1351, opened 2023-10-13
Prevent annotations from vanishing when toggling or canceling #1321

### LINKED ISSUE #1321: Sample shape disappears if edit state initialized and no edits made
Sample Taking & Analysis Area Workflows:
1. Sample info or Region Step, respectively: draw & save n>=2 features
2. Click to initiate edit state on one feature, but do not make any edits
3. Move focus to the other feature
       Result: first feature "vanishes". Refreshing the page will restore it.

https://github.com/archesproject/arches-for-science/assets/22161487/ad77c9a9-acac-4f5c-9886-d230e1397105

jacobtylerwalls:
Fixes #1321

Before, upon creating n>=2 analysis (or sample) area annotations, two scenarios caused a shape to vanish from the canvas (but not from the sidebar).

1 - Click twice on feature 1 to get edit controls. Deselect it. Result: feature 1 disappears

2 - With feature 2 selected, click Cancel.
Result: feature 2 disappears

Fixing this almost caused a regression in this workflow, so it should be tested also:
1. Save shape 1
2. Save shape 2
3. Select shape 1, make edits, save.
4. Select shape 2
5. Select shape 1 (Result: shape 1 morphs into the first saved state, not the second.)

This was avoided by skipping a tile.reset() call. Worth some thought whether that's okay.

Another issue was almost caused when deleting annotations, so that's worth testing too.

### DIFF
--- a/arches_for_science/media/js/views/components/workflows/analysis-areas-workflow/analysis-areas-annotation-step.js
+++ b/arches_for_science/media/js/views/components/workflows/analysis-areas-workflow/analysis-areas-annotation-step.js
@@ -428,8 +428,13 @@ define([
                 if(response.ok){
                     return;
                 }
-                
-                throw response;
+                response.json().then(function(error){
+                    params.pageVm.alert(new params.form.AlertViewModel(
+                        "ep-alert-red",
+                        error.title,
+                        error.message,
+                    ));
+                });
             }).then(function(data){
                 parentPhysicalThing.data[physicalThingPartAnnotationNodeId].features().forEach(function(feature){
                     self.deleteFeature(feature);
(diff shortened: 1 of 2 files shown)


chiatt (commented):
The changes in Analysis areas look good and are working as expected. I see the regression you mentioned about geometry changes reverting when switching between edited features, but your changes are still a big improvement. For sample taking I'm getting an error when I try to create a sample: 
image
and when I delete a sample I get this error: 
image
Do you see the same thing?

jacobtylerwalls:
Thanks, I'll look closer on Monday to see where the analysis and sample workflows are diverging.

jacobtylerwalls:
I'm going to try a new approach; this one is dead-ending.

PR #1353, opened 2023-10-13
Disable save and cancel buttons in sample/analysis workflows #1347

### LINKED ISSUE #1347: Analysis area: save and cancel buttons styled as disabled aren't disabled
The save and complete buttons when correctly styled as disabled are not disabled in fact. (The Save button will still fire off a request, and the cancel button will still reset the canvas.)

Found while dev'ing #1321

jacobtylerwalls:
Before, the save & cancel buttons were styled as disabled but not disabled in fact.

Fixes #1347


njkim (approved):
It looks good and works as expected.

PR #1357, opened 2023-10-13
Fix highlighting of resource instances #1356

### LINKED ISSUE #1356: Physical thing search results do not indicate if in collection
In the 'Add Objects to Your Project' step of the 'Update Project Collection Workflow' if a user selects a physical thing to add to a collection, it gets listed on the right. However, there is no indication in the list on the left that it has been added. That means that somebody could remove an object from a project by accident if they did not see that it was already added on the right. 

If an object has already been added to the collection there should be a clear indication that it has already been added, or it should hidden entirely from the search results so that it can only be removed from the collection by clicking the remove button in the list on the right.
image

jacobtylerwalls:
Closes #1356
Regression in 3e3bd2a.

(Missed this in the template when refactoring the form values in the viewmodel to be shaped like resource objects, rather than lists of resource ids.)


chiatt (approved):
Looks good!

PR #1358, opened 2023-10-16
Handle when there is no custom renderer, #1352

### LINKED ISSUE #1352: Deleting an ascii file reader config does not remove config from XY Loader UI
If a user deletes a reader config from the UI, the config persists even though it was successfully deleted on the backend. Refreshing the page reveals the deletion was successful

njkim:
Handle the case when there is not custom render for xy upload workflow.
Fix the issue with UI not updating when the renderer is deleted, #1352


chiatt (approved):
Looks good!

PR #1361, opened 2023-10-17
Remove temp_file view from afs

njkim:
Remove the temp_file view from arches-for-science, re: move the temp_file view to core, archesproject/arches#9749


chiatt (approved):
Looks good!

PR #1366, opened 2023-10-17
Add migration granting Resource Reviewers view plugin access

### LINKED ISSUE #10064: Need user group permissioning for access to workflows (dev/1.0.x)
To access the workflows, it seems like one needs to be a superuser AND member of all (or some?) of the Arches predefined groups. We should probably create a "Workflow Group" to grant access to certain users to access the workflows. 

(cc @annabelleee who may have more ideas)

jacobtylerwalls:
Closes archesproject/arches#10064

Grant resource reviewers access to plugins on an object-level for Plugin objects for maximal flexibility.

A data migration is provided to act on existing installed plugins, and a post_sql statement is included with the package to handle future installations (for which this data migration would be pointless, since they lack any plugins before loading packages).

### Testing instructions ###
- [ ] migrate on an otherwise migrated application
- [ ] reverse migrate
- [ ] setup_db
- [ ] load package, ensure post_sql runs
- [ ] Functional test: give `tester3` the Resource Reviewer role, ensure they can access workflows


arches_for_science/migrations/0001_initial.py
```diff
@@ -8,6 +8,7 @@ class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
+        ("models", "10121_workflowhistory"),
```
jacobtylerwalls, line 11: Open to other ways of doing this, but without this I was finding:
```py
django.db.utils.IntegrityError: null value in column "name" of relation "django_content_type" violates not-null constraint
DETAIL:  Failing row contains (1, null, models, bulkindexqueue).
```

aarongundel: When testing, I didn't see this error after loading from scratch.  I went back to zero, removed the dependencies, and ran them forward and the migration applied without issue.

aarongundel: Issue is that core is referenced in the second migration.  There's a load order issue that can occur when loading from scratch.  This solution seems good to me.

aarongundel (approved):
LGTM

jacobtylerwalls:
Thanks for the review!

PR #1367, opened 2023-10-18
Update preferred terms concept, #1330

### LINKED ISSUE #1330: Add preferred terms to Name Type - Physical Thing collection

njkim:
Update the preferred terms concept and the related collections, #1330 

1. Move the concept to Name Type - Generic
2. Add the concept to Name Type - Generic and Name Type - Physical Thing


chiatt (approved):
Looks good!

PR #1374, opened 2023-10-19
Add error handling to project name step #1354

### LINKED ISSUE #1354: Completing a step after resolving a TileValidationError jumps over a step
New project workflow:
1. Fill out scope, leave name blank
2. Next
3. Fill out name
4. Next

Expected: navigated to step 2
Actual: navigated to step 3

https://github.com/archesproject/arches-for-science/assets/38668450/d6829dd0-7675-48f5-9d3c-2e31f7a331c5

jacobtylerwalls:
Closes #1354 (well, in a way...)

**Before**
In #1354, the video shows a TileValidationError causing the new project workflow to skip from step 1 to step 3 (or as many steps as errors you caused). Since #1340 was merged, to reproduce it, you need to throw a TileValidationError or some other backend error somewhere in tile.py.

**After**
Now, by setting params.form.error(), we ensure that the promise is rejected. Here is the subscription to `params.form.error`, where it runs `componentBasedStepReject`:

https://github.com/archesproject/arches/blob/d084723fd602d81cc88f357162878fc27e5a069c/arches/app/media/js/views/components/workflows/workflow-component-abstract.js#L811-L816

**Next steps**
Maybe we should audit the other workflow steps for this pattern?

### DIFF
--- a/arches_for_science/media/js/views/components/workflows/create-project-workflow/project-name-step.js
+++ b/arches_for_science/media/js/views/components/workflows/create-project-workflow/project-name-step.js
@@ -83,7 +83,7 @@ define([
                         new params.form.AlertViewModel('ep-alert-red', data.title, data.message)) 
                     })
                     .catch(_ => { params.pageVm.alert(
-                        new params.form.AlertViewModel('ep-alert-red', 'Error saving project name')) 
+                        new params.form.AlertViewModel('ep-alert-red', params.form.error() || 'Error saving project name')) 
                     });
                 }
             });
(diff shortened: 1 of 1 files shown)


chiatt (commented):
This looks good, but I'm wondering if the error that ultimately appears on line 86 should be `params.form.error()`

jacobtylerwalls:
Good catch, that could make a difference if (the last value of) self.error was actually `Error saving project type`, for instance.

chiatt (approved):
Looks good!

Issue #1373, opened 2023-10-19
Installation README Ambiguity (dev/1.1.x)


ekansa:
In the installation README (https://github.com/archesproject/arches-for-science/blob/dev/1.1.x/README.md#2-project-configuration), there seems to be some ambiguity about the 'myappname' app added to the `INSTALLED_APPS` config in `settings.py`. 

Should this be removed or updated?

ekansa:
The project settings.py that I created from dev/7.5.x doesn't have `RENDERERS` defined. Maybe we should add something like this?

```python
# A hack to see if this exists.
try:
    RENDERERS
except NameError:
    RENDERERS = []

RENDERERS += [
    {
        "name": "xy-reader",
        "title": "XY Data File Reader",
        "description": "Use for all instrument outputs with x-y data",
        "id": "e93b7b27-40d8-4141-996e-e59ff08742f3",
        "iconclass": "fa fa-bolt",
        "component": "views/components/cards/file-renderers/xy-reader",
        "ext": "txt",
        "type": "text/plain",
        "exclude": "",
    },
]

```

ekansa:
Also an issue with the changes to urls.py:

```
from django.conf.urls import path, include
arches                | ImportError: cannot import name 'path' from 'django.conf.urls' (/root/.local/lib/python3.10/site-packages/django/conf/urls/__init__.py)
```

ekansa:
Hmmm...

Maybe a Django version issue with this:
```
from django.conf.urls import path, include
ImportError: cannot import name 'path' from 'django.conf.urls' (/root/.local/lib/python3.10/site-packages/django/conf/urls/__init__.py)
```

Something like this work?

```python
from django.urls import include, path, re_path
from django.conf import settings
from django.conf.urls.static import static


urlpatterns = [
    re_path(r'^', include('arches.urls')),
    re_path(r"^", include("arches_for_science.urls")),
    path("reports/", include("arches_templating.urls")),
] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```

jacobtylerwalls:
Thanks, these are really helpful observations.

> The project settings.py that I created from dev/7.5.x doesn't have RENDERERS defined. 

[...]

PR #1379, opened 2023-10-24
Add required `FORMATS` setting to README

jacobtylerwalls:
Removed from docs in d82bb80, but still depended on here:

https://github.com/archesproject/arches-for-science/blob/14de8d020428b4bb7ee719d63c327629164aa3fe/arches_for_science/utils/context_processors.py#L26

Reported by @ekansa in #1373

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,18 @@ RENDERERS += [
 INSTRUMENT_FILE_FORMATS = ["txt"]
 
 X_FRAME_OPTIONS = "SAMEORIGIN"
+
+FORMATS = [
+    {"name": "Bruker M6 (point)", "id": "bm6", "renderer": "31be40ae-dbe6-4f41-9c13-1964d7d17042"},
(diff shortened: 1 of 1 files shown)


ekansa:
@jacobtylerwalls Thanks! This seems to have done the trick. I could get AfS started successfully in my Docker setup

chiatt:
Sorry - I thought we'd already removed FORMATS from any afs code when I removed that setting. We actually don't need that setting any longer, so I'm going to close this and address the issue here: #1381

ekansa:
@chiatt that makes sense not to hardcode FORMATS in the settings.py 

I raised the issue because, without the FORMATS, my AfS deployment didn't work. So a better fix would be welcome

jacobtylerwalls:
@ekansa I just merged #1381, which removes the functionality of the `FORMATS` setting entirely, so once you've pulled from arches_for_science, you should be good to remove your `FORMATS` setting, which is just cruft now.

ekansa:
@jacobtylerwalls Thank you! I'll remove the `FORMATS` from my Docker deployment too.

PR #1383, opened 2023-10-25
Bump `arches-templating` to 0.1.2

jacobtylerwalls:


jacobtylerwalls:
Actually, I just noticed that we already had a `>=` requirement, so this isn't very urgent. But still worth clarifying that 0.1.1 is not recommended (could pull in a bad 3rd party dep)

aarongundel (approved):
LGTM

PR #1384, opened 2023-10-25
fixes for #1328 and #1382

### LINKED ISSUE #1328: Show parsing error if delimiter fails to parse in 2D loader
For example if the delimiter is '**'

### LINKED ISSUE #1382: Warn a user that an xy config is in use on deletion
Currently a user can delete an xy renderer config in the xy upload step regardless of whether that config is in use. Users should be warned or prevented from deleting configs in use.

aarongundel:
fixes for #1328 and #1382


chiatt (changes requested):
The error notification indicating that a config is already in use activates as expected, but I can't close it. I think it'd be ok to use the page banner alert if that's easier.

arches_for_science/templates/views/components/plugins/importer-configuration.htm
```diff
@@ -3,6 +3,7 @@
     <h4><span>{% trans "Importer Configurations" %}</span></h4>
     <span>{% trans "Define how to import your x/y data" %}</span>
     <div>
+        <div class="afs-error-message ep-alert-red" style="padding:5px;" data-bind="visible: rendererInUse">{% trans "This importer is in use - it should be replace it in all files where it is used before deleting." %}</div>
```
chiatt: Looks like a typo in the error text. Maybe 'replace it' should just be 'replaced'

chiatt (changes requested):

chiatt (approved):

PR #1386, opened 2023-10-26
Persist the selected file - file interpretation step, #1376

### LINKED ISSUE #1376: Return to previously selected file in file interpretation step
When a user switches between files in the file interpretation step by selecting the 'Show Files List' button, the previous file selection is cleared. Instead, it should remain selected until a user selects a different file.

image
image

njkim:
Persist the selected file between toggles, #1376 
the file interpretation step in the xy upload workflow


chiatt (approved):
Looks good!

PR #1395, opened 2023-10-27
Show physical thing icons in resource select widgets #1391

### LINKED ISSUE #1391: Prevent question mark from appearing in resource instance dropdown
Physical things are represented with a ? in the first step of the chemical analysis workflow. In the project report workflow, the icon is shown, and in upload xy, it's blank. It would be good if the icon was shown, but if that's difficult to do nothing should be shown. What's important is that it isn't a question mark. 

image
image
image

jacobtylerwalls:
Fixes #1391


chiatt (approved):
lgtm!

PR #1401, opened 2023-10-30
allow for configurable columns

### LINKED ISSUE #1387: Users need to be able to define xy column names in config
When defining a configuration in the xy loader, users need to be able to define the x and y column names for display in the plot. Currently these are hard-coded in the xy-reader as 
```      
this.chartTitle("XRF Spectrum");
this.xAxisLabel("keV");
this.yAxisLabel("Count");
```

aarongundel:
Allows for configurable column/chart title in the xy loader file interpretation step


chiatt (approved):
Looks good!

PR #1402, opened 2023-10-30
Removes elements from annotation screenshot

### LINKED ISSUE #1399: Remove elements outside of annotation screenshot
Currently if an annotation screenshot is collected in the reporting workflow, the canvas thumbnail, title, and list of parts is included in the screenshot. Instead, only the leaflet map should be included, ideally without the navigation buttons in the upper left.

aarongundel:
Clips out extraneous elements from leaflet map in screenshot - just screenshots the leaflet map without navigation.


chiatt (approved):
Looks good!

PR #1406, opened 2023-10-31
Fix sample deletion #1405

### LINKED ISSUE #1405: Deleting a sample or analysis doesn't clear it from the canvas right away
Regression in 3a9a29ef4aa4c2e0a1f84611a4a6cbeb52f6557d.

jacobtylerwalls:
Before, deleting a sample that is selected (e.g. because it had just been saved) would take effect on the tile but not clear itself from the canvas.

***Testing instructions***
- [ ] deletion of samples/analyses takes effect on the canvas immediately
- [ ] saving, manipulating still works, no shapes get respawned onto the canvas out of nowhere

Closes #1405


chiatt (approved):
Looks good!

PR #1413, opened 2023-11-01
1134 add chemical analysis image workflow

chrabyrd:
Adds workflow for adding derivative images of chemical analyses


chiatt (approved):
There are a few issues that came up when testing: #1420, #1421, and #1422. They prevented me from getting all the way through the workflow, but I think we should merge this in and fix those issues afterward.

PR #1415, opened 2023-11-01
Update project download, #1398

### LINKED ISSUE #1398: Add template and annotation screenshots to file download
When a user goes through the reporting workflow, they select screenshots, then files, then create the report. All of these files should be included with the report in the final step.

njkim:
Let the user to download all the selected files at the end of the workflows. #1398

### DIFF
--- a/arches_for_science/tasks.py
+++ b/arches_for_science/tasks.py
@@ -13,17 +13,17 @@ except ImportError:
 
 
 @shared_task
-def download_project_files_task(userid, files, project_name):
+def download_project_files_task(userid, files, temp_files, project_name):
     from arches_for_science.views.download_project_files import FileDownloader
 
     downloader = FileDownloader()
     user = User.objects.get(pk=userid)
-    exportid, files = downloader.create_download_zipfile(user, files, project_name)
+    exportid, files = downloader.create_download_zipfile(user, files, temp_files, project_name)
 
     search_history_obj = SearchExportHistory.objects.get(pk=exportid) if exportid else None
 
     msg = _(
-        "The related files for the project '{}' are ready for download. You have 24 hours to download your files before they are automatically removed."
+        "The report(s) and the related file(s) for the project '{}' are ready to download. You have 24 hours to download your files before they are automatically removed."
     ).format(project_name)
     notiftype_name = "Search Export Download Ready"
     context = dict(
--- a/arches_for_science/tasks.py
+++ b/arches_for_science/tasks.py
@@ -23,7 +23,7 @@ def download_project_files_task(userid, files, temp_files, project_name):
     search_history_obj = SearchExportHistory.objects.get(pk=exportid) if exportid else None
 
     msg = _(
-        "The report(s) and the related file(s) for the project '{}' are ready to download. You have 24 hours to download your files before they are automatically removed."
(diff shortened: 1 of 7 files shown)


arches_for_science/tasks.py
```diff
@@ -13,17 +13,17 @@
...
-    exportid, files = downloader.create_download_zipfile(user, files, project_name)
+    exportid, files = downloader.create_download_zipfile(user, files, temp_files, project_name)
 
     search_history_obj = SearchExportHistory.objects.get(pk=exportid) if exportid else None
 
     msg = _(
-        "The related files for the project '{}' are ready for download. You have 24 hours to download your files before they are automatically removed."
+        "The report(s) and the related file(s) for the project '{}' are ready to download. You have 24 hours to download your files before they are automatically removed."
     ).format(project_name)
     notiftype_name = "Search Export Download Ready"
```
chiatt, line 28: Can you change this name to just _('Download Ready')?

arches_for_science/tasks.py
```diff
@@ -13,17 +13,17 @@
...
     downloader = FileDownloader()
     user = User.objects.get(pk=userid)
-    exportid, files = downloader.create_download_zipfile(user, files, project_name)
+    exportid, files = downloader.create_download_zipfile(user, files, temp_files, project_name)
 
     search_history_obj = SearchExportHistory.objects.get(pk=exportid) if exportid else None
 
     msg = _(
-        "The related files for the project '{}' are ready for download. You have 24 hours to download your files before they are automatically removed."
+        "The report(s) and the related file(s) for the project '{}' are ready to download. You have 24 hours to download your files before they are automatically removed."
```
chiatt: I don't think we can assume that the files will be removed in 24 hours. Could this just be shortened to: `The report(s) and the related file(s) for the project, '{}' are ready for download.` BTW, I think both 'for' and 'to' work, but your use of 'for' in your earlier commit sounds a bit better to me.

arches_for_science/views/download_project_files.py
```diff
@@ -40,17 +71,24 @@ def post(self, request):
...
+        for file in all_files:
+            for file_object in all_file_objects:
+                if str(file_object.fileid) == str(file["fileid"]):
                     file["file"] = file_object.path
         download_files = []
         skipped_files = []
         size_limit = 104857600  # 100MByte
-        for file in files:
+        for file in all_files:
             if file["file"].size >= size_limit:
```
chiatt: I discovered that this will break if the file cannot be found. Could you wrap this if/else in a try block and log an error message? Something like: logger.warn(_("Unable to locate {}".format(file["file"]))?

chiatt (changes requested):
This is looking good. Just a few minor changes and it should be ready to go.

arches_for_science/tasks.py
njkim: I believe this is the notification type name in the db.
https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/models/migrations/5613_notification_type.py#L67
and gets called here
https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/tasks.py#L360

chiatt: Okay, thanks. Never mind that change then.

chiatt (approved):

PR #1418, opened 2023-11-02
Chemical analysis workflow Final step table, #1414

### LINKED ISSUE #1414: Remove file int and params from last step in chem upload workflow
There's no file interpretation step, so these values will never be entered in the workflow

image

njkim:
Remove columns with no data during the workflow.
- interpretation, parameter


chiatt (approved):
Looks good!

PR #1427, opened 2023-11-06
Fix sample/analysis name validation

### LINKED ISSUE #1419: User gets no error when trying to save a sample without a required node
When saving a sample, a user must provide a name, otherwise the save will fail. When it does, the user should get an alert notifying them that they need to provide a sample name. 

image

jacobtylerwalls:
- Fix client-side validation of sample and analysis names
- Surface server errors that might get through anyway

Closes #1419

The fact that empty key-value pairs like `{ value: '', direction: '' }` are truthy was causing issues.


aarongundel (approved):
Looks good!  Works for me.

Issue #1424, opened 2023-11-03
Update select/clear all buttons in file selection step in reporting workflow


chiatt:
The selection buttons at the top of the file list should be changed to match the select/clear buttons in the xy file loader.

PR #1431, opened 2023-11-08
Update file select buttons in report wf, #1424

njkim:
Update the select buttons in the file select step, report workflow.


chiatt (approved):
Looks good!

Issue #1449, opened 2023-11-17
Add observation type input to workflows


chiatt:
In the chem analysis data upload and upload xy data workflows we need to add a concept dropdown allowing users to select an observation type.

PR #1450, opened 2023-11-20
adds observation type input to workflow steps #1449

chrabyrd:


chiatt (changes requested):
This looks good; it adds the observation type tile as expected. However, if a user advances to the next step, and then returns to the observation step and changes the value, the change is not reflected in the tile.

chrabyrd:
> This looks good; it adds the observation type tile as expected. However, if a user advances to the next step, and then returns to the observation step and changes the value, the change is not reflected in the tile.

@chiatt This is symptomatic of a larger issue. If a user completes the step then goes back and changes a value in the step, instead of the change being applied to the tile, a second tile is created. I will fix the observation tiles/steps as part of this ticket but we should file a follow-up ticket to check/correct all steps for this behavior.

chiatt:
> > This looks good; it adds the observation type tile as expected. However, if a user advances to the next step, and then returns to the observation step and changes the value, the change is not reflected in the tile.
> 
> @chiatt This is symptomatic of a larger issue. If a user completes the step then goes back and changes a value in the step, instead of the change being applied to the tile, a second tile is created. I will fix the observation tiles/steps as part of this ticket but we should file a follow-up ticket to check/correct all steps for this behavior.

https://github.com/archesproject/arches-for-science/issues/1451

chrabyrd:
> This looks good; it adds the observation type tile as expected. However, if a user advances to the next step, and then returns to the observation step and changes the value, the change is not reflected in the tile.

@chiatt resolved in latest 👍

chiatt (approved):

PR #1452, opened 2023-11-21
Add multi-card primary descriptor #1368

### LINKED ISSUE #1368: Usability enhancement: Display provenance of samples and analysis areas in search and dropdowns
We ran into an interesting issue where there are repeated names for certain paintings (physical things), and those names are used to name samples and analysis areas that derive from those paintings. This creates confusion for users, because without additional identifying information (such as accession numbers), it is hard to know which of the multiple "Profile of a Woman" paintings a given sample derived from.

We should explore how to carry forward accession numbers or other identifiers from the physical thing (of study) down to samples and analysis areas. That way samples and analysis areas can be more easily traced back to their parent physical thing in the user interface.

Perhaps the graph can be updated to include a branch for something like "Provenance Description" that a workflow can automatically populate with names and identifiers for parent physical objects when a user defines a new sample or analysis area instance?

jacobtylerwalls:
### Description ###
Retires the primary descriptor from core arches that was limited to a single card. Creates a new primary descriptor that can draw from any card (using node aliases, which are unique), and wires it into the existing location for managing this (the function manager).

A migration deploys triggers to the tiles table that will calculate descriptors and writes to the resource instance table. (Another trigger captures changes to the function config itself and recalculates every resource.)

This migration was generated by `django-pgtrigger`, which abstracts away some of the boilerplate surrounding the trigger (e.g `FOR EACH ROW` etc), and allows for Django-y filters like `Q()` to add conditions on the trigger. This trigger is particularly complicated, but I can see this being helpful for simpler ones. Another benefit is managing the triggers in the application code and getting simple git diffs on the source.

### Setup instructions ###
- depends on archesproject/arches#10286 to register the function
- depends on archesproject/arches#10300 to pass down the descriptor type
- depends on archesproject/arches#10303 to fix migration order failure
- in testing project, e.g. disco, add to settings like so:
```py
FUNCTION_LOCATIONS.append('arches_for_science.pkg.extensions.functions')
```
- reinstall arches-for-science dependencies to get `django-pgtrigger`
- reload the afs package in testing project (e.g. disco)
- run migrations

Fixes #1368


arches_for_science/pkg/extensions/functions/multicard_resource_descriptor.py
```diff
@@ -0,0 +1,42 @@
...
+}
+
+
+class MulticardResourceDescriptor(AbstractPrimaryDescriptorsFunction):
+    """Implemented in the database via triggers on tiles table.
+    
+    This implementation just fetches the calculated result from the db."""
+
+    def get_primary_descriptor_from_nodes(self, resource, config, context=None, descriptor="name"):
+        # TODO: patch core arches to send the appropriate descriptor argument, e.g. "name" or "description" etc.
```
jacobtylerwalls: @chiatt We should consider patching this in Arches 7.5. As it stands, this function has no knowledge of what descriptor was passed in (which we need if we want to be able to read the JSON field from the db and look up the given descriptor by key).

arches_for_science/pkg/extensions/functions/multicard_resource_descriptor.py
```diff
@@ -0,0 +1,42 @@
...
+        result = ""
+        requested_language = context.get("language", None) if context else None
+
+        lookup_language = requested_language or get_language() or settings.LANGUAGE_CODE
+        try:
+            result = resource.descriptors[lookup_language][descriptor]
```
bferguso, line 47: @jacobtylerwalls  - Would this work instead? That way we don't need to add the descriptor type to the call signature.
```
result = resource.descriptors[lookup_language][config["type"]]
```

jacobtylerwalls: Thanks @bferguso, that's an interesting idea that I'm open to.

My thought at the moment is that adding a `type` key to the config object is duplicative, though, and potentially more work to implement.

I happen to be adding a new descriptor function in this PR (so I'm on the hook for writing a data migration anyway), but suppose I weren't -- I'd need to write a data migration to reshape the config objects to provide that missing key, when all of the knowledge is already in the python file.

arches_for_science/templates/views/components/functions/multicard-resource-descriptor.htm
```diff
@@ -0,0 +1,167 @@
...
+
+                <div class="col-xs-12">
+                    <input 
+                        data-bind="select2Query: {
+                            select2Config: {
+                                ...baseSelect2Config,
+                                value: selectedNodes.name,
+                            }
+                        }"
+                    ></input>
```
jacobtylerwalls: Re #1454 I know I'll at least have to do this once archesproject/arches#10265 lands, but this wasn't enough to get the dropdown working when merging the accessibility branch. It may be because I'm using the `data` prop?
```suggestion
                    <select 
                        data-bind="select2Query: {
                            select2Config: {
                                ...baseSelect2Config,
                                value: selectedNodes.name,
                            }
                        }"
                    ></select>
```

arches_for_science/media/js/views/components/functions/multicard-resource-descriptor.js
```diff
@@ -0,0 +1,95 @@
...
+                            self.updateTemplate(action.value, action.status, observableName)
+                        })
+                    }, this, 'arrayChange')
+                }
+            );
+
+            this.baseSelect2Config = {
+                multiple: true,
+                placeholder: arches.translations.selectPrimaryDescriptionIdentifierCard,
+                data: {results: self.groupedNodesForSelect2},
```
jacobtylerwalls: When the accessibility work is merged I'll confer with @apeters about how to get this static data loaded after the updates to the select2 binding.

arches_for_science/media/js/views/components/functions/multicard-resource-descriptor.js
```diff
@@ -0,0 +1,95 @@
...
+                map_popup: ko.observableArray(
+                    self.parseNodeIdsFromStringTemplate(self.map_popup.string_template())
+                ),
+            };
+
+            this.groupedNodesForSelect2 = self.graph.cards.map(card => {
+                return {
+                    text: card.name,
+                    children: self.graph.nodes.filter(
+                        node => node.datatype === 'string' && node.nodegroup_id === card.nodegroup_id
```
jacobtylerwalls: Here is the reason for doing the grouping with `children`: we can get nodes grouped under their card.

arches_for_science/media/js/views/components/functions/multicard-resource-descriptor.js
```diff
@@ -0,0 +1,95 @@
...
+                    ).map(node => {
+                        return {
+                            id: node.nodeid,
+                            text: node.alias,
+                        }
+                    }),
+                }
+            });
```
jacobtylerwalls, line 74: @chiatt What order should the cards be in in the dropdown?

arches_for_science/migrations/0005_deploy_triggers_for_primary_descriptor.py
```diff
@@ -0,0 +1,66 @@
...
...y)::text, '\\\"')\n                END\n                INTO working_string;\n\n                SELECT REPLACE(\n                    working_string,\n                    alias_with_separators,\n                    TRIM((inner18n_obj_for_lang -> 'value')::text, '\"')\n                )\n                INTO resolved_node_value_for_lang;\n\n                -- Update the working value\n                SELECT jsonb_set(\n                    localized_result_all,\n                    ARRAY[lang, descriptor_key],\n                    TO_JSONB(resolved_node_value_for_lang)\n                ) INTO localized_result_all;\n\n                IF descriptor_key = 'name' THEN\n                    SELECT jsonb_set(\n                        localized_result_name_only,\n                        ARRAY[lang],\n                        TO_JSONB(resolved_node_value_for_lang)\n                    ) INTO localized_result_name_only;\n                END IF;\n\n            END LOOP;\n\n        END LOOP;\n\n    END LOOP;\n\n    UPDATE resource_instances\n    SET descriptors = localized_result_all, name = localized_result_name_only\n    WHERE resourceinstanceid = resourceid;\n\nEND LOOP;\nEND;\nRETURN NULL;\n",
```
jacobtylerwalls: The function dump (for this particularly complicated function) is ugly, but it's not a migration we'll be manually editing.

arches_for_science/pkg/graphs/resource_models/Group.json
```diff
@@ -7837,20 +7837,20 @@
                         "descriptor_types": {
                             "description": {
                                 "nodegroup_id": "32db94f8-c05e-11e9-a2c1-a4d18cec433a",
-                                "string_template": "<Statement_content>"
+                                "string_template": "<statement_content>"
```
jacobtylerwalls, line 7840: These now need to be node aliases rather than node names. I could make the lookup case-insensitive, I suppose?

arches_for_science/trigger_functions.py
```diff
@@ -0,0 +1,184 @@
...
+    nodegroup_for_alias UUID;
+    localized_string_node_value jsonb;
+    lang text;
+    inner18n_obj_for_lang jsonb;
+    resolved_node_value_for_lang text;
+    working_string text;
+    localized_result_all jsonb;
+    localized_result_name_only jsonb;
+
+BEGIN
```
jacobtylerwalls, line 28: Here's where we'll get git diffs/blames all in one place for trigger logic, instead of only spread across multiple migration files.

arches_for_science/trigger_functions.py
```diff
@@ -0,0 +1,184 @@
...
+IF FOUND THEN
+    SELECT JSONB_OBJECT('{}') INTO localized_result_all;
+    SELECT JSONB_OBJECT('{}') INTO localized_result_name_only;
+    SELECT fn_config -> 'descriptor_types' -> 'name' -> 'string_template' INTO name_template;
+    SELECT fn_config -> 'descriptor_types' -> 'map_popup' -> 'string_template' INTO map_popup_template;
+    SELECT fn_config -> 'descriptor_types' -> 'description' -> 'string_template' INTO description_template;
+
+    SELECT ARRAY[name_template, map_popup_template, description_template] INTO all_templates;
+    SELECT 0 INTO loop_index;
+    FOR this_template IN SELECT UNNEST(all_templates)
```
jacobtylerwalls, line 65: Overall, the approach is:
```
For each descriptor...
    For each node alias in the template...
        For each language in the i18n obj for that node...
            Update the template string to replace that specific node
```
I'm doing string replacements, as I go, e.g.
`Prefix <value1> Middle Value <value2> Suffix`
becomes
`Prefix ResolvedValue1 Middle Value <value2> Suffix`
and then
`Prefix ResolvedValue1 Middle Value ResolvedValue2 Suffix`
.

This involves a good deal of casting text (on which I did the string replacement) to JSON (so I can store the intermediate results). It just seemed to me that if I created a variable to hold the all the localized strings _as strings (and not JSON)_, then I'd be creating a nested tree very similar (identical) to the one I could just update as I go, and this would make the code harder to read. I'm open to ideas here.

arches_for_science/trigger_functions.py
```diff
@@ -0,0 +1,184 @@
...
+                    ELSE TRIM((localized_result_all -> lang -> descriptor_key)::text, '\"')
+                END
+                INTO working_string;
+
+                SELECT REPLACE(
+                    working_string,
+                    alias_with_separators,
+                    TRIM((inner18n_obj_for_lang -> 'value')::text, '"')
+                )
+                INTO resolved_node_value_for_lang;
```
jacobtylerwalls, line 169: What should the fallback value be when there is no value for the given node? Leave `Prefix <statement_content> Suffix`, which is at least somewhat informative (and what happens now), or just `Prefix Suffix`?

When it's just that a language is missing on the node vlalue being worked on, currently the fallback is replace the alias with nothing/empty string (like the latter option just mentioned). At the moment this is just a happy accident of the way languages are stored as `{"de": "value": "" ...}` instead of `{"de": "value": null, ...}`.

arches_for_science/trigger_functions.py
```diff
@@ -0,0 +1,184 @@
...
+    UPDATE resource_instances
+    SET descriptors = localized_result_all, name = localized_result_name_only
+    WHERE resourceinstanceid = resourceid;
+
+END IF;
+END;
+RETURN NULL;
+"""
+
+#### Make two adjustments to create a very similar function that works on ALL resource instances.
```
jacobtylerwalls, line 134: I wasn't thrilled about this string replacement to create the second trigger, but only two small places need to be edited, and this avoids duplicating a lot of code.

I considered f-strings, but I got nervous about dealing with escaping curly brackets. I'm open to it, though.

arches_for_science/trigger_functions.py
jacobtylerwalls: Updated in [13f025d](https://github.com/archesproject/arches-for-science/pull/1452/commits/13f025da92ae955a169a138f3267e7cf772d96f3) to fall back to ` -- ` when a value for a node is missing.

Missing languages still just fall back to whitespace.

README.md
```diff
@@ -28,12 +28,13 @@ If you don't already have an Arches project, you'll need to create one by follow
 Since Arches for Science uses `Cantaloupe` as its IIIF server, take notice of the
 Cantaloupe [installation instructions](https://arches.readthedocs.io/en/stable/developing/advanced/managing-and-hosting-iiif/), too.
 
-When your project is ready add "arches_templating" and "arches_for_science" to INSTALLED_APPS and "arches_for_science" to ARCHES_APPLICATIONS in your project's settings.py file:
+When your project is ready add "arches_templating", "arches_for_science", and "pgtrigger" to INSTALLED_APPS and "arches_for_science" to ARCHES_APPLICATIONS in your project's settings.py file:
```
chiatt, line 31: Could you also add instruction to add `FUNCTION_LOCATIONS.append('arches_for_science.pkg.extensions.functions')`?

arches_for_science/media/js/views/components/functions/multicard-resource-descriptor.js
chiatt: I think alphabetically would be great

arches_for_science/pkg/graphs/resource_models/Group.json
chiatt: That would work in many cases, but a node alias can differ from the node name in other ways, generally because an alias must be unique to a graph and a name only must be unique to its siblings. So, I think it's good to be explicit about changing from names to aliases as you've done here.

arches_for_science/pkg/graphs/resource_models/Physical Thing.json
```diff
@@ -8618,15 +8618,15 @@
                     "config": {
                         "dateFormat": "YYYY-MM-DD",
                         "defaultValue": "",
-                        "label": "Lateset Begin",
+                        "label": "Latest Begin",
```
chiatt, line 8621: Nice catch!

arches_for_science/migrations/0005_deploy_triggers_for_primary_descriptor.py
```diff
@@ -0,0 +1,65 @@
...
+            options={
+                "proxy": True,
+                "indexes": [],
+                "constraints": [],
+            },
+            bases=("models.tilemodel",),
+        ),
+        pgtrigger.migrations.AddTrigger(
+            model_name="functionxgraphproxy",
+            trigger=pgtrigger.compiler.Trigger(
```
chiatt, line 37: I would expect the trigger to be removed on the reverse migration, but it persists. When you reverse migrate, are the pgtriggers dropped?

jacobtylerwalls: Works for me:

Reversing from 0008 to 0004:
```sh
root@d6b16aa0f3ba:/web_root/disco# python3 manage.py migrate arches_for_science 0004
System check identified some issues:

WARNINGS:
?: (urls.W005) URL namespace 'admin' isn't unique. You may not be able to reverse all URLs in this namespace
?: (urls.W005) URL namespace 'oauth2' isn't unique. You may not be able to reverse all URLs in this namespace
Operations to perform:
  Target specific migration: 0004_improve_primary_descriptor, from arches_for_science
Running migrations:
  Rendering model states... DONE
  Unapplying arches_for_science.0008_remove_tilemodelproxy_calculate_multicard_primary_descriptor_single_and_more... OK
  Unapplying arches_for_science.0007_update_primary_description_trigger_handle_static_value... OK
  Unapplying arches_for_science.0006_update_primary_description_trigger_null_handling... OK
  Unapplying arches_for_science.0005_deploy_triggers_for_primary_descriptor... OK
root@d6b16aa0f3ba:/web_root/disco# python3 manage.py dbshell
psql (14.9 (Ubuntu 14.9-1.pgdg22.04+1), server 14.5 (Debian 14.5-1.pgdg110+1))
Type "help" for help.

disco=# \dS tiles
...
Triggers:
    __arches_check_excess_tiles_trigger BEFORE INSERT ON tiles FOR EACH ROW EXECUTE FUNCTION __arches_check_excess_tiles_trigger_function()
    __arches_trg_update_spatial_attributes AFTER INSERT OR DELETE OR UPDATE ON tiles DEFERRABLE INITIALLY DEFERRED FOR EACH ROW EXECUTE FUNCTION __arches_trg_fnc_update_spatial_attributes()
```

I noticed that there was a missing migration on this branch from Friday, that might have been the issue.

arches_for_science/migrations/0005_deploy_triggers_for_primary_descriptor.py
```diff
@@ -58,7 +58,6 @@ class Migration(migrations.Migration):
                     operation="INSERT OR UPDATE OR DELETE",
                     pgid="pgtrigger_calculate_multicard_primary_descriptor_single_26494",
                     table="tiles",
-                    timing="DEFERRABLE INITIALLY DEFERRED",
```
jacobtylerwalls: We should just be careful to remake migrations instead of trying to edit the files; luckily CI caught the drift between the trigger defs and the calculated state.

jacobtylerwalls:
@chiatt FYI -- during release testing I just discovered that node aliases are nullable. So this hasn't really been tested against that (I would guess that which node we resolve `<>` to is probably undefined).

chiatt:
> @chiatt FYI -- during release testing I just discovered that node aliases are nullable. So this hasn't really been tested against that (I would guess that which node we resolve `<>` to is probably undefined).

There must have been a reason for making them nullable when the alias was introduced. I don't recall what that would have been, but perhaps it should be changed - in 7.6 at least.

chiatt (approved):
Looks great!

PR #1453, opened 2023-11-22
Enhances upload failures

### LINKED ISSUE #1448: Failing chem analysis upload
When uploading a chem analysis file in the Chem Analysis Data Upload workflow, the file appears to have uploaded, but the upload fails:

image

In s3, the file also appears to have uploaded, but the size of the file (in this case 5MB) shows 0 bytes.
image

aarongundel:
Makes it more obvious when uploads fail due to CORS permissions and such.  Also removes the file tile.


chiatt (approved):
Looks good!

PR #1459, opened 2023-11-30
selectwoo fixes for accessibility

apeters:
a bunch of fixes to the dropdowns used in AFS to support the accessibility work in core Arches

this requires that all the accessibility work has been merged into Arches

to test, all dropdowns should work as expected

### DIFF
--- a/arches_for_science/media/js/views/components/plugins/file-configuration.js
+++ b/arches_for_science/media/js/views/components/plugins/file-configuration.js
@@ -8,8 +8,7 @@ define([
     'templates/views/components/plugins/file-configuration.htm',
     'bindings/codemirror',
     'views/components/cards/file-renderers/xy-reader',
-    'bindings/select2v4',
-    'select2'
+    'bindings/select2-query'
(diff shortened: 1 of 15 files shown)


chiatt (changes requested):
This is looking much better, but I noticed that users can now create physical things in the 'select-physical-thing' step. This is a problem because if they were to do that, the new physical thing would not be in their selected project and that would be bad. 

image

Another issue is that I can't select a sampler in the sample taking workflow. If I try, the input remains blank, but I get an error when I try to save the workflow and then can't proceed.

Screenshot 2023-11-30 at 4 35 49 PM

apeters:
> This is looking much better, but I noticed that users can now create physical things in the 'select-physical-thing' step. This is a problem because if they were to do that, the new physical thing would not be in their selected project and that would be bad.
> 
> image
> 
> Another issue is that I can't select a sampler in the sample taking workflow. If I try, the input remains blank, but I get an error when I try to save the workflow and then can't proceed.
> 
> Screenshot 2023-11-30 at 4 35 49 PM

fixed in a4a020e

also relies on updates to the accessibility branch

chiatt (approved):

PR #1462, opened 2023-12-01
Remove step info from workflows #1456

### LINKED ISSUE #1456: Remove step information from workflows
AFS will rely on the workflow/plugin help rather than the step information defined in each workflow. For this ticket we just need to remove the informationboxinfo from each step in each of the workflow files. eg: https://github.com/archesproject/arches-for-science/blob/b8ec57078aabf843214d1e539a4965e2e5c00f98/arches_for_science/media/js/views/components/plugins/sample-taking-workflow.js#L24-L27

jacobtylerwalls:
Closes #1456


chrabyrd (approved):
lgtm!

Issue #1451, opened 2023-11-21
Ensure updating tiles in workflow steps does not result in duplicate tiles


chiatt:
In some cases saving a tile in a step, advancing to the next step and then returning to the previous step to make an edit can result in a second tile rather than a change to the first. This issue was found in the Upload XY data step, but we need to go through our workflows and ensure it's not happening elsewhere.

https://github.com/archesproject/arches-for-science/pull/1450#issuecomment-1821554063

chiatt:
@njkim did not observe this happening in other parts of the workflows

PR #1473, opened 2023-12-04
Improve undo action in instrument info steps #1428

### LINKED ISSUE #1428: [observation info] Cannot write value to a ko.computed unless you specify `write`
X/Y Workflow, observation info:
- Select an instrument
- Undo

```js
Uncaught (in promise) Error: Cannot write a value to a ko.computed unless you specify a 'write' option. If you wish to read the current value, don't pass any parameters.
    at WorkflowComponentAbstract.e [as hasUnsavedData] (knockout-latest.js:54:426)
    at params.form.reset (instrument-info-step.js:176:19)
    at WorkflowComponentAbstract._resetComponent (workflow-component-abstract.js:677:38)
    at eval (workflow-step.js:161:39)
    at new Promise (<anonymous>)
    at eval (workflow-step.js:160:30)
    at Array.forEach (<anonymous>)
    at eval (workflow-step.js:159:63)
    at new Promise (<anonymous>)
    at WorkflowStep.undo (workflow-step.js:157:14)
```

jacobtylerwalls:
Fixes #1428

- Fixes "cannot write to ko.computed"
- Fixes Undo button still showing active after using it to clear
- Allows clearing date entries with Undo

- Fixes observation type cannot be cleared with undo button (#1475)

### DIFF
--- a/arches_for_science/media/js/views/components/workflows/chemical-analysis-workflow/ca-instrument-info-step.js
+++ b/arches_for_science/media/js/views/components/workflows/chemical-analysis-workflow/ca-instrument-info-step.js
@@ -49,7 +49,7 @@ define([
         let procedureTileId = getProp('procedure', 'tileid');
         let projectTileId = getProp('project', 'tileid');
         let observedThingTileid = getProp('observedThing', 'tileid');
-        let observationTypeTileId = getProp('type', 'tileid');
+        let observationTypeTileId = getProp('observationType', 'tileid');
         let dateTileId = getProp('date', 'tileid');
         let nameTileId = getProp('name', 'tileid');
 
@@ -67,6 +67,7 @@ define([
 
         const snapshot = {
             dateValue: self.dateValue(),
+            observationType: self.observationType(),
             instrumentValue: self.instrumentValue(),
             procedureValue: self.procedureValue(),
             parameterValue: self.parameterValue()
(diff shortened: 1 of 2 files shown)


arches_for_science/media/js/views/components/workflows/chemical-analysis-workflow/ca-instrument-info-step.js
```diff
@@ -165,10 +169,12 @@ define([
         };
 
         params.form.reset = function(){
+            self.nameValue(null);
+            self.instrumentName(null);
```
jacobtylerwalls: These needed to be cleared out because of the subscriptions that get called e.g. instrumentValue() right after this. If there is still any data here, it was preventing getting the form state to be exactly like the original empty state.

jacobtylerwalls: Actually, I found a better place to do this (inside the instrumentName subscription). The first version here was too aggressively setting to null.

arches_for_science/media/js/views/components/workflows/chemical-analysis-workflow/ca-instrument-info-step.js
```diff
@@ -49,7 +49,7 @@ define([
         let procedureTileId = getProp('procedure', 'tileid');
         let projectTileId = getProp('project', 'tileid');
         let observedThingTileid = getProp('observedThing', 'tileid');
-        let observationTypeTileId = getProp('type', 'tileid');
+        let observationTypeTileId = getProp('observationType', 'tileid');
```
jacobtylerwalls, line 52: @njkim I'm wondering if this was the source of the extra tile from #1451 🤔 ?

aarongundel (approved):
Tested, works great.

PR #1476, opened 2023-12-04
file interpretation step fix

### LINKED ISSUE #1430: [file interpretation step] Save & cancel styled as disabled but aren't
Found this by tabbing to what I thought was the next field and pressing enter; instead the Save button executed even though it was styled as disabled.

Similar to #1347 but for file interpretation step in X/Y loader.

aarongundel:
fixes file interpretation save - saving when it's disabled.


jacobtylerwalls (approved):
Tested, works well

PR #1481, opened 2023-12-04
Fix button toggling when manipulating sample areas #1471

### LINKED ISSUE #1471: Toggle sample buttons
Add new sample and return to sample list buttons should not appear at the same time, but instead should toggle as they do for the analysis areas workflow.

image

jacobtylerwalls:
Closes #1471 (port same solution from analysis workflow)


chiatt (approved):
Looks good to me!

PR #1483, opened 2023-12-05
Moves selection into observations

### LINKED ISSUE #1457: Move select/deselect all in the reporting file selection step into each observation
In the reporting workflow file selection step a user can select/clear all the files in all the observations. This should be moved into the header of each observation so that it only applies to the respective observation dataset.

aarongundel:
Replaces global [select all] with an observation scoped [select all]

closes #1457


chiatt (approved):
Looks good!

PR #1513, opened 2024-01-02
Fetch descriptors from database #1511

### LINKED ISSUE #1511: Descriptors not set when loading business data
Load afs-test-pkg, which creates resources and tiles.

Check descriptors on resource instances: still show Undefined. Would have expected descriptors to be calculated after #1452. Maybe something's overwriting it?

jacobtylerwalls:
Before, importing business data was overwriting the correctly calculated descriptors with Undefined because we were providing the same python object to `save_descriptors()`, oblivious to what had been calculated in the db trigger.

***Testing instructions***
```sh
python3 manage.py packages -o load_package -a arches_for_science -db -dev
python3 manage.py packages -o load_package -s <path to afs-test-pkg/pkg>
```

Then check physical things, ensure they have descriptions. Best to check this with Celery. It worked without celery (probably because we were refetching resource instances at some point).


jacobtylerwalls:
To test, you will also need to delete the multicard_resource_descriptor.py from DISCO if you already had one, due to https://github.com/archesproject/arches/issues/10492.

chiatt (approved):

PR #1515, opened 2024-01-04
Mark strings for localization throughout the project #1503

### LINKED ISSUE #1269: Handle missing required values using server response
> I see this follows the same pattern as the `Project Name` node, but we should actually handle for missing required values using the response from the server.  That way the error message will be properly localized. We can address this for both nodes in another PR. This fix prevents the step from breaking, though, so it's a good stop-gap.

_Originally posted by @chiatt in https://github.com/archesproject/arches-for-science-prj/pull/1262#discussion_r1291905713_

### LINKED ISSUE #1503: Localize Arches for Science
Ensure all strings in the project can be localized

### LINKED ISSUE #1504: Fix default name for analysis areas in the Analysis Area workflow
The default name should begin with 'Analysis Area' rather than 'Region'. Reported by VGM

### LINKED ISSUE #1505: Ensure the users see the word 'image service' rather than 'manifest'
We have replaced the term 'manifest' with 'image service' because it is easier for users unfamiliar with iiif to understand. There are still cases - in error messages for example - were the word 'manifest' is still used. We should ensure that it is replaced by 'image service'. Reported by VGM

### LINKED ISSUE #1517: [i18n] check if some strings are user-facing
Identified on localization branch with `appears in UI?`

jacobtylerwalls:
**Changes**
- Mark strings for localization throughout the project.
- Fix #1269
- Follow-up to #1262, looks like some strings were missed there
- [Rename "region" to "analysis area"](https://github.com/archesproject/arches-for-science/pull/1515/commits/bc66efe164eb4e77eeeb02766a7058db98301358) https://github.com/archesproject/arches-for-science/issues/1504
- Identify follow-ups with `TODO(i18n)`:
  - #1516
  - #1518
  - #1519
  - #1520

There's a start toward the last item in b452d6bc17deb842a0358b1c8ff87034b9648d68.

**Other stuff:**
- Require arches 7.6 (since this PR assumes archesproject/arches#10498 is going in)
- Add LOCALE_PATHS to settings (so that .po files can be generated)
   - To look at the .po files:
```
cd arches-for-science/arches_for_science
PYTHONPATH='../../arches' python3 ../manage.py makemessages --local=nl --settings='arches_for_science.test_settings' --extension=htm,py
```
The PYTHONPATH is because arches.tests is not importable otherwise.

cd'ing to the app directory makes it easier to ignore build/ and node_modules/ files without having to specify that on the CLI, but it produces slightly different output than running from the project dir, so I'll defer to @aarongundel on how we're going to do that in prod. But this works for conjuring up a .po file to look at. I don't know if we're only going to generate files for disco, or if we're going to generate them for AfS and commit to the AfS repo.

- PR contains #1513 fix, but will rebase that out later when it's merged on its own
- I took the liberty of standardizing a couple report table headers and error messages where the slight discrepancies would have created unnecessary work for translators IMO.
- Since I was on the hook 🪝 for combing through every file anyway and checking that aria-labels were marked for translation, I took the liberty of adding `aria-hidden="true"` on decorative icons. [More at MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden).

**Testing instructions:**
It's too much to test, but sanity checking a few workflows would be good.
Make sure you check out archesproject/arches#10498 before testing.
To reset your db:
```
python3 manage.py packages -o load_package -a arches_for_science -db -dev
python3 manage.py packages -o load_package -s <path to afs-test-pkg/pkg>
```
You might also enable `'nl'` in LANGUAGES in disco/settings.py

You could also generate the .po files as demo'd above and glance over the strings to ensure they would make sense to a translator and have punctuation (if punctuated in English).

Closes #1503

**Notes**
There are probably some unnecessary strings in the .po files that we could get from the node labels, but I didn't want to tackle that refactor now when we have more important follow-ups to get to.


jacobtylerwalls:
Oh, I guess we have a management command for fetching PO files, but haven't tested yet if it works with an arches app like this. It looks like it also handles graph/card/widget labels.

arches_for_science/templates/javascript.htm
```diff
@@ -31,6 +31,131 @@
...
+    deleting='{% trans "Deleting" as deleting %} "{{ deleting|escapejs }}"'
+    loading-datasets='{% trans "Loading datasets..." as loadingDatasets %} "{{ loadingDatasets|escapejs }}"'
+    loading-samples-analyses='{% trans "Loading samples and analysis areas..." as loadingSamplesAnalyses %} "{{ loadingSamplesAnalyses|escapejs }}"'
+    saving='{% trans "Saving" as saving %} "{{ saving|escapejs }}"'
+    saved='{% trans "Saved." as saved %} "{{ saved|escapejs }}"'
+    takes-time='{% trans "This may take a while, please do not move away from this step." as takesTime %} "{{ takesTime|escapejs }}"'
+    takes-too-long='{% trans "This is taking longer than usual. Thank you for your patience." as takesTooLong %} "{{ takesTooLong|escapejs }}"'
+    analysis-area='{% trans "Analysis Area" as analysisArea %} "{{ analysisArea|escapejs }}"'
+    analysis-areas='"{{ analysisAreas|pluralize }}"'
+    no-modifiying-areas='{% trans "Analysis Areas may not be modified in the sample taking workflow" as noModifyingAreas %} "{{ noModifyingAreas|escapejs }}"'
```
jacobtylerwalls:
```suggestion
    no-modifying-areas='{% trans "Analysis Areas may not be modified in the sample taking workflow" as noModifyingAreas %} "{{ noModifyingAreas|escapejs }}"'
```

chiatt (commented):
I went through each of the workflows and nearly everything worked as expected. Just a few things came up:

The step names aren't appearing in the Chem Analysis Data workflow
image

I'm not able to move to the last step of the review dataset workflow because arches is undefined
image

I ran into an unhandled promise error in the template selection step of the reporting workflow. I don't know why it would be an issue with this branch, but switching back to the dev/1.1.x branch, the step worked okay. I'm using the template in the `document_templates` directory in disco.

arches_for_science/media/js/views/components/workflows/project-report-workflow/report-template-select.js
```diff
@@ -19,7 +19,7 @@ define([
         this.getTemplates = async() => {
             const response = await fetch(arches.urls.reports_list);
 
-            if(response.ok) {
+            if (!response.ok) {
```
chiatt: I think this should remain `if(response.ok)`. It looks like this might be causing the unhandled promise error.

chiatt (approved):
Looks good to me!

Issue #1525, opened 2024-01-04
[i18n] Several workflows assume text entries are made in the active language


jacobtylerwalls:
_Sample/Analysis location steps_

If a user picks a language in the rich text widget language switcher, it may not be the same as the active language, and so the user's choice will be discarded:

https://github.com/archesproject/arches-for-science/blob/600c7e0516ff894008826ef78297e7701e53040f/afs/media/js/views/components/workflows/sample-taking-workflow/sample-taking-sample-location-step.js#L609-L610

(others in this file, too)

We should decide whether to support this use case or disable the language switcher somehow.

jacobtylerwalls:
Same in upload-dataset:
```js
part.displayname = part.data[physicalThingPartNameNodeId][arches.activeLanguage]['value'];
```

Issue #1506, opened 2023-12-21
Add Dutch to the list of available languages in the VGM project and enable the language switcher


chiatt:

jacobtylerwalls:
@aarongundel I figure you may want to do what I did here: 98c09bc4b1be0c23844bcc46bb53f528fe7079cd

PR #1526, opened 2024-01-04
Add Dutch language concept to workflows #1506

jacobtylerwalls:
Refs #1506

Some of these values should probably come from the text widgets instead, but that's a question posed on #1525.


aarongundel:
holding per feedback from @chiatt

Issue #1540, opened 2024-01-17
Update observation types to divide types into 2d/3d


chiatt:
Observation Type:

- 2D Data Measurement
    - FORS Measurement (Alt label*: Fiber Optics Reflectance spectroscopy (FORS) Measurement)
    - FTIR Measurement (Alt label*: Fourier-transform infrared (FTIR) spectroscopy Measurement)
    - Point Raman Measurement (Alt label*: Point Raman spectroscopy Measurement)
    - Point XRF Measurement (Alt label*: Point X-ray fluorescence spectroscopy (XRF) Measurement)
- 3D Data Measurement
    - 3D XRF Measurement (Alt label*: 3D X-ray fluorescence (XRF) spectroscopy Measurement)
    - RIS Measurement (Alt label*: Reflectance imaging spectroscopy (RIS) Measurement)
    - SEM-EDS Measurement (Alt label*: Scanning electron microscopy – energy dispersive (SEM-EDS) spectroscopy Measurement)
    - 3D Raman Measurement (Alt label*: 3D Raman spectroscopy Measurement)

PR #1541, opened 2024-01-17
Add migration to update obs types, #1540

chiatt:
Updates observation types with 2d and 3d grouping. When QAing, confirm that the observation type dropdown looks like this:
image


aarongundel (approved):
Tested package load and it worked.  Rolled back migration and forward without issue.  LGTM!

PR #1543, opened 2024-01-18
Reindex digital resources on manifest save

### LINKED ISSUE #1529: IIIF Services Digital Resources are not reindexed when after manifest changes
To reproduce:

1. Create a manifest with a canvas
2. Notice that the primary descriptor for the digital resource is the name of the first uploaded canvas
3. Change the name of the manifest and save
4. Notice that the name of the digital resource instance has changed, but the indexed primary descriptor hasn't.

chiatt:
Reindexes digital resources when a user changes a manifest's name or description.

To test:
Create a manifest - look in search and check that a digital resource exists for your manifest
Change the name and/or description in the manifest manager
Revisit search and the descriptors should be updated

### DIFF
--- a/arches_for_science/utils/digital_resource_for_manifest.py
+++ b/arches_for_science/utils/digital_resource_for_manifest.py
@@ -80,7 +80,7 @@ def add_tiles(
         else:
             return
         name_tile.data[name_content_node_id] = build_string_object(name)
-        name_tile.save(transaction_id=transactionid, index=False)
+        name_tile.save(transaction_id=transactionid, index=True)
 
     if statement:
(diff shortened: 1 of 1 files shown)


jacobtylerwalls (approved):
+1 tested.

Some things I noticed that are for future:
- We index the resource twice this way. Would be nice to do it once, but I guess that comes at the cost of having to fetch the Resource proxy model again, given that the Tile proxy model doesn't save a reference to it. Doing that would be a nice core arches improvement.
- Doing that could make this future-proof against changing the descriptor to read from other tiles, e.g. type or service
- Indexing seems to happen irrespective of whether the transaction succeeds or is rolled back. Do we have a story around what happens to the index if the transaction fails?

PR #1544, opened 2024-01-18
Update version and arches req version

chiatt:
Update alpha version and make arches 7.6.0a1 the min arches version


aarongundel (approved):
LGTM

PR #1552, opened 2024-02-13
Avoid lazy strings in initial migration

jacobtylerwalls:
Lazy strings make sense for user-facing help texts, etc., but not when running initial migrations to save a single hard-coded name.


aarongundel (approved):
LGTM

PR #1553, opened 2024-02-13
upgrade the config for s3 to use sigv4 for uppy endpoint

aarongundel:
Change the uppy config to use sigv4 by default.  Uppy doesn't respect django storages signature version by default, so we need to manually enforce it.

### DIFF
--- a/arches_for_science/views/s3.py
+++ b/arches_for_science/views/s3.py
@@ -1,4 +1,3 @@
-
 import json
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


chiatt (commented):
It looks like this is failing to run because it is missing the 'AWS_S3_REGION_NAME' setting.

jacobtylerwalls:
Yeah just need to update the test_settings.py

chiatt (approved):
LGTM!

PR #1557, opened 2024-02-15
Lint backend

jacobtylerwalls:
Add a CI step to lint the backend. Discussed with @aarongundel after the avoidable error in #1551.

#### Alternatives considered
pylint -- would have caught #1551, deeply configurable
flake8 -- focused more on style/cosmetics. wasn't immediately clear how to just filter down on the important stuff.
ruff -- the new hotness in python linting, but I still don't think it would have caught #1551. sure it will one day, though, and we can switch over when it's fast/mature.

### DIFF
--- a/arches_for_science/views/analysis_area_and_sample_taking.py
+++ b/arches_for_science/views/analysis_area_and_sample_taking.py
@@ -336,6 +336,8 @@ class SaveSampleAreaView(SaveAnnotationView):
         )
         tile.save(transaction_id=transactionid, request=request, index=False)
 
+        return tile
+
     def post(self, request):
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
See failed action on first commit 👍

arches_for_science/views/analysis_area_and_sample_taking.py
```diff
@@ -336,6 +336,8 @@ def save_removed_from_tile(self, request, sample_resourceid, removed_from_resour
         )
         tile.save(transaction_id=transactionid, request=request, index=False)
 
+        return tile
```
chiatt, line 339: Oh - interesting. It caught a potentially serious bug.

chiatt (approved):
Looks good to me!

PR #1560, opened 2024-02-28
Dev/1.1.x Workflow in-app help

ekansa:
This PR has updates to the in-app help for AfS workflows. They appear to render well, but I'd check them again, since I'm having some styling issues now that AfS shifted to use Arches dev/7.6.x .


chiatt (approved):
This looks great. We will need to wrap all of this text into trans blocks, but we can address that in a separate PR.

Issue #1547, opened 2024-01-25
If s3 isn't available, dependent workflows are hidden, but not heading


chiatt:
If s3 isn't available, dependent workflows are hidden, but not heading in init-workflow plugin

<img width="2007" alt="image" src="https://github.com/archesproject/arches-for-science/assets/1877663/fedf663e-9eaf-4dc4-b6b6-4287115dc3fd">

PR #1563, opened 2024-03-13
Prevent display of heading when related workflows are not visible #1547

chiatt:
Prevent display of heading when related workflows are not visible because project is not using s3 #1547


aarongundel (approved):
LGTM

PR #1565, opened 2024-03-27
Anticipate v8 breaking change to TileModel.nodegroup

jacobtylerwalls:
As part of reviewing archesproject/arches#9613 I ran the AfS test suite, and found these breaking changes impacted.

We change change nodegroup -> nodegroup_id already in v7.6 without waiting for v8. This will make it easier to develop against either version.

~Setting to DRAFT: there could be more that aren't covered by tests; need to grep the codebase.~

Also add DATABASES to test_settings so that it doesn't attempt to use the core arches test db. Might segregate out this change if this gets fixed in the app templates first.


chiatt (approved):
Looks good!

PR #1570, opened 2024-06-26
upgrade to core 7.6.x

aarongundel:


jacobtylerwalls:
Just checking it's on the radar to run updateproject also?

chiatt:
> Just checking it's on the radar to run updateproject also?

arches-for-science doesn't have a webpack directory, so I don't think that running updateproject is really necessary in this case, but I could be overlooking something.

jacobtylerwalls:
It copies over a bunch of other goodies like pre-commit, vue, and typescript tooling--at least the pre-commit tooling will be nice if you want to adopt black in this project.

But we might want to wait for a decision on archesproject/arches#11016, too.

jacobtylerwalls (changes requested):
We should update `PhysicalThingSearchView.get_permitted_nodegroups` to drop the ".pk"

arches_for_science/media/js/views/components/workflows/sample-taking-workflow/sampling-info-step.js
```diff
@@ -85,7 +85,9 @@ define([
         });
         
         this.samplingDate.subscribe(function(val){
-            self.samplingName(arches.translations.samplingActivityOf.replace('{physicalThingName}', self.physicalThingNameValue) + ' ' + val);
+            const name = arches.translations.samplingActivityOf.replace('{physicalThingName}', self.physicalThingNameValue) + ' ' + val;
+            self.samplingName()[arches.activeLanguage].value = name
+            self.samplingName(self.samplingName());
```
jacobtylerwalls: Do we have the same problem in the two places from instrument-info-step.js in [this commit](https://github.com/jacobtylerwalls/arches-for-science-prj/commit/696415b5f86c7dcd2afbdae0b025ceea03bc81e1)?

chiatt:
It looks like we will have to add a settings.py file - at least to run arches commands:

```
 % python manage.py updateproject

No Django settings specified.
Unknown command: 'updateproject'. Did you mean startproject?
Type 'manage.py help' for usage.
```

jacobtylerwalls:
You might be able to pass the test_settings file to --settings.

arches_for_science/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class ArchesForScienceConfig(AppConfig):
+    name = "arches_for_science"
+    verbose_name = "Arches_for_science"
```
jacobtylerwalls: This is the human readable name, so we can:
```suggestion
    verbose_name = "Arches for Science"
```

arches_for_science/media/js/views/components/workflows/sample-taking-workflow/sampling-info-step.js
chiatt: Thanks - good catch

chiatt: It turns out that instrument-info-step only has the string - rather than the localized object in the counter-part functions, so we don't have to make this change in that step.

tests/test_settings.py
```diff
@@ -0,0 +1,120 @@
...
+except ImportError:  # unable to import prior to installing requirements
+    pass
+
+APP_ROOT = os.path.dirname(__file__)
+TEST_ROOT = os.path.normpath(os.path.join(ROOT_DIR, "..", "tests"))
+
+
+MIN_ARCHES_VERSION = arches.__version__
+MAX_ARCHES_VERSION = arches.__version__
+
```
jacobtylerwalls, line 35:
```suggestion
```

pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-  "arches>=7.6.0a1,<8",
+  "arches @ git+https://github.com/archesproject/arches.git@dev/7.6.x",
```
jacobtylerwalls: Maybe we should loosen this a little bit.
```suggestion
  "arches>=7.6.2,<8",
```

jacobtylerwalls (approved):

PR #1573, opened 2024-06-27
Fix used-before-assignment in ManifestXCanvasView

jacobtylerwalls:
Fix CI failure on [recent pushes](https://github.com/archesproject/arches-for-science/actions/runs/9672816483/job/26685757036)


chiatt (approved):
Looks good!

PR #1575, opened 2024-07-03
Accept uuids rather than nodegroup objects

chiatt:
Fix permission check to handle for 7.6.x changes to permission interface


aarongundel (approved):
Funny - I'd already made this locally, just not committed yet.  LGTM.

PR #1576, opened 2024-07-18
Update app to run as a project

chiatt:
Updates arches for science to run both as an Arches App and independent project. Includes project updates to support latest changes in Arches dev/7.6.x and moves tests under project's test directory


.github/workflows/main.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - name: Check for missing migrations
         working-directory: afs
         run: |
-          PYTHONPATH="../arches" python manage.py makemigrations --check --settings=arches_for_science.test_settings
+          PYTHONPATH="../arches" python manage.py makemigrations --check --settings=tests.test_settings
```
jacobtylerwalls: Let's remove the path manipulation workaround -- it was only here because the previous test settings file was much smaller and just imported test settings from arches.tests.test_settings.

Also, let's just use the normal settings file now.

```suggestion
          python manage.py makemigrations --check
```

.github/workflows/main.yml
```diff
@@ -74,4 +74,4 @@ jobs:
       - name: Run afs unit tests
         working-directory: afs
         run: |
-          PYTHONPATH="../arches" python -W default::DeprecationWarning manage.py test --settings=arches_for_science.test_settings
+          PYTHONPATH="../arches" python -W default::DeprecationWarning manage.py test tests --settings=tests.test_settings
```
jacobtylerwalls:
```suggestion
          python -W default::DeprecationWarning manage.py test tests --settings=tests.test_settings
```

arches_for_science/test_settings.py
```diff
@@ -1,39 +0,0 @@
...
-    "django_celery_results",
-    "arches_for_science",
-    "pgtrigger",
-    # "debug_toolbar"
-)
-ROOT_URLCONF = "arches_for_science.urls"
-
-LOCALE_PATHS = [os.path.join(APP_ROOT, "locale")]
-
-FUNCTION_LOCATIONS = ["arches_for_science.pkg.extensions.functions"]
```
jacobtylerwalls, line 35: Noticed that `LOCALE_PATHS` and `FUNCTION_LOCATIONS`[1] are different here than in the new test settings file. Consider having the new test settings file do this first:
```
from arches_for_science.settings import *
```

Do you want that change in the project template?
***

[1] `FUNCTION_LOCATIONS` might be moot after archesproject/arches#10613 and archesproject/arches#11016, so not a big deal, mostly just interested in the right principle of settings inheritance.

webpack/webpack-utils/build-filepath-lookup.js
```diff
@@ -0,0 +1,46 @@
...
+            {
+                fileList.push(childPath);
+            }
+            return fileList;
+            }, []);
+        };
+
+    return getFileList(path).reduce((lookup, file) => {
+        const extension = file.match(/[^.]+$/).toString();
+        const extensionReplacementRegex = new RegExp(`\\.${extension}$`);
```
jacobtylerwalls, line 34: Now or later we should rerun updateproject to get the fix for dotfiles.

tests/functions/multicard_resource_descriptor.py
```diff
@@ -0,0 +1,57 @@
+from arches.app.functions.primary_descriptors import AbstractPrimaryDescriptorsFunction
```
jacobtylerwalls: Why did this file get created? If it got created automatically with package load or updateproject or some other command, that seems like a serious arches bug to look into before 7.6 final.

tests/base_test.py
```diff
@@ -0,0 +1,79 @@
...
+        # Unless the user has something other than the Django default, give them
+        # what they probably want.
+        if kwargs["pattern"] == "test*.py":
+            kwargs["pattern"] = "*.py"
+        super().__init__(*args, **kwargs)
+
+    def setup_databases(self, **kwargs):
+        ret = super().setup_databases(**kwargs)
+
+        app_settings()  # adds languages to system
```
jacobtylerwalls: Let's remember to apply the fix from archesproject/arches#11205

arches_for_science/settings.py
```diff
@@ -0,0 +1,398 @@
...
+
+
+WEBPACK_LOADER = {
+    "DEFAULT": {
+        "STATS_FILE": os.path.join(APP_ROOT, '..', 'webpack/webpack-stats.json'),
+    },
+}
+
+DATATYPE_LOCATIONS.append('arches_for_science.datatypes')
+FUNCTION_LOCATIONS.append('arches_for_science.functions')
```
jacobtylerwalls: Maybe this is follow-up work, but it seems like we should be able to get rid of the function file under pkg/extensions if we have this now. (Maybe the situation improved in 7.6 after archesproject/arches#10613?)

jacobtylerwalls (changes requested):

webpack/webpack-utils/build-filepath-lookup.js
chiatt: done!

arches_for_science/test_settings.py
chiatt: I've removed this file

tests/functions/multicard_resource_descriptor.py
chiatt: I'm really not sure how it ended up in there (maybe just an errant copy and paste?), but I've removed it.

arches_for_science/settings.py
chiatt: done!

tests/base_test.py
jacobtylerwalls: I'll manually reconcile this (it doesn't get updated by updateproject)

jacobtylerwalls (approved):

.github/workflows/main.yml
```diff
@@ -1,77 +1,272 @@
...
-          POSTGRES_DB: arches
+          POSTGRES_DB: arches_for_science
         ports:
           - 5432:5432
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12"]
```
jacobtylerwalls, line 25: Since Arches for Science is conceived primarily as a library, I guess I'd prefer to see the full range of python versions it supports in the test matrix. (I don't know why the project settings template was changed in this respect, as it didn't go through code review.)

jacobtylerwalls: never mind! this is now build target branch only, false alarm!

package.json
```diff
@@ -29,6 +29,8 @@
         "eslint:watch": "nodemon --watch . --ext ts,vue --exec npm run --silent eslint:check",
         "gettext:extract": "vue-gettext-extract",
         "gettext:compile": "vue-gettext-compile",
+        "prettier:check": "prettier glass/src --check",
+        "prettier:fix": "prettier glass/src --write",
```
jacobtylerwalls:
```suggestion
        "prettier:check": "prettier arches_for_science/src --check",
        "prettier:fix": "prettier arches_for_science/src --write",
```

pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
```
jacobtylerwalls, line 28: The github action expects there to be an optional dependency block here, so this also needs reconciling with the template.

jacobtylerwalls:
The line ending check was added because it bit Brett during development. But we probably don't need to care about ontology files. Easy enough to fix here, but do you think this will be enough of a pain point for the community that we should exempt .xml files?

chiatt:
> The line ending check was added because it bit Brett during development. But we probably don't need to care about ontology files. Easy enough to fix here, but do you think this will be enough of a pain point for the community that we should exempt .xml files?

Ah - I was wondering why that was there. It's hard to say how much trouble it might be for people, but I'd be in favor of the exemption just in case.

PR #1578, opened 2024-08-29
Pin uppy version

chiatt:
Pins uppy version because latest version breaks webpack build


aarongundel:
@chiatt should this be passing now?

aarongundel (approved):
LGTM

PR #1581, opened 2024-08-30
Use arches test runner instead of base_test

chiatt:
Use arches shared test runner instead of base_test. see archesproject/arches/pull/11359


aarongundel (approved):
LGTM

PR #1588, opened 2024-12-03
Ensure part tiles are removed post deletion

### LINKED ISSUE #1589: Deleting an analysis area or sample area causes workflows to fail
Deleting a physical thing that represents an analysis area or sample area causes workflows to fail. This is because the creation of these objects results in a 'parts' tile on the parent resource. If either of these types of resources is deleted, the parts tile persists on the parent. In several places throughout the workflows, the parts tiles are used to inspect their corresponding objects. The workflows will fail at these points if the tile's target resource does not exist in the system.

chiatt:
Ensure part tiles are removed post deletion


aarongundel (approved):
LGTM

aarongundel (approved):
LGTM x 2!

PR #1591, opened 2024-12-04
Updates instructions for installing or upgrading

chiatt:
Updates instructions for installing or upgrading

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,13 @@ Navigate to the `arches-for-science` directory from your terminal and run:
 ```
 pip install -e .
 ```
+`Important`: Installing the arches-for-science app will install Arches as a dependency. This may replace your current install of Arches with a version from PyPi. If you've installed Arches for development using the `--editable -e` flag, you'll need to re pip install arches after installing arches-for-science
 
 ### If installing for deployment, run:
 ```
 pip install arches-for-science
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,3 +5,114 @@ Arches for Science an Arches Application designed to help conservation scientist
...
+
+    XY_TEXT_FILE_FORMATS = ["txt"]
+
+    X_FRAME_OPTIONS = "SAMEORIGIN"
+    ```
+
+4. Next ensure arches and arches_for_science are included as dependencies in package.json
+    ```
+    "dependencies": {
+        "arches": "archesproject/arches#dev/7.6.x",
```
aarongundel, line 90: Should we note that in production installs people should probably point to stable branches?

README.md
```diff
@@ -5,3 +5,114 @@ Arches for Science an Arches Application designed to help conservation scientist
...
+```
+
+### If installing for deployment, run:
+```
+pip install arches-for-science
+```
+
+`Important`: Installing the arches-for-science app will install Arches as a dependency. This may replace your current install of Arches with a version from PyPi. If you've installed Arches for development using the `--editable` flag, you'll need to re pip install arches after installing arches-for-science
+
+## Project Configuration
```
aarongundel, line 28: This seems mostly geared toward development.  Perhaps in another iteration we can make some documentation for a production release.

aarongundel (approved):
LGTM

README.md
chiatt: Good point. I've moved that text up to the 'Installing for development section', but I agree we should provide more details on installing for prod.

PR #1593, opened 2024-12-04
Update install instructions with required js dependencies

chiatt:
Update install instructions with required js dependencies


aarongundel (approved):
LGTM

PR #1601, opened 2025-01-14
Modify renderer path

### LINKED ISSUE #1602: Error when loading xy-renderer
An error occurs when loading the xy-renderer because the endpoint to get the renderer is the config endpoint rather than the renderer endpoint.

chiatt:
Modifies the path to get the renderer rather than a specific config


aarongundel (approved):
LGTM

PR #1605, opened 2025-02-12
Address bugs in sample taking and analysis area steps

chiatt:
- Allows user to scroll side panel in sample taking step
- Prevents signal receiver from attempting to delete tiles that do not exist (causing error)
- Prevents user from attempting to save an analysis area without an annotation geometry
- Removes asterisks from analysis and sample taking steps because there is already a 'required' label and the asterisk was obscuring the label

### DIFF
--- a/arches_for_science/media/css/project.css
+++ b/arches_for_science/media/css/project.css
@@ -561,6 +561,14 @@
 
 }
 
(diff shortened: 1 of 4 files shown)


aarongundel (approved):
LGTM

PR #1608, opened 2025-02-25
Return thumbnails for manifest image digital resources

### LINKED ISSUE #1607: 500 error when fetching some thumbnails
Thumbnail fetcher is getting an image url rather than json causing a 500 error when loading search results. The error is occurring on the Tate instance here: `arches_for_science/utils/thumbnail_fetchers/digital_resource_thumbnail_fetcher.py, line 22, in get_thumbnail`

chiatt:
If a digital resource is an image that belongs to a manifest rather than the manifest itself, it's identifier is a link to the image rather than the UUID of the manifest. The thumbnail fetcher was implemented before images belonging to manifests were represented as digital resources, so the fetcher was not written to handle a link. This PR no allows both a link or a manifest uuid to return a thumbnail image.

Closes #1607


aarongundel (approved):
LGTM

PR #1611, opened 2025-03-18
Add project and django request logging

jacobtylerwalls:
Refs archesproject/arches#11861

AfS logs in a few places, but no loggers were listening (no "arches_for_science" logger configured).

Disco should probably make the same change.


aarongundel (approved):
lgtm

PR #1615, opened 2025-06-09
Update graphs with export configuration

### LINKED ISSUE #1612: Add data export info to the models
Change the models so that nodes for export are specified by enabling the "Export via Search Results" toggle and adding a shapefile fieldname.

chiatt:
I'm thinking this could go in a patch because it won't affect any existing instances, only new ones.


aarongundel (approved):
LGTM

Issue #1614, opened 2025-06-09
Add support for v3 of the iiif presentation api


chiatt:
Arches currently only supports v2 of iiif presentation api. However, many iiif servers are now using v3. We can't add a new feature in Arches until v8.1 early next year, so in the interim this work would need to be in AFS until it can be moved into core. Or, more likely, iiif could be moved out of core and maintained as an app with presentation api v3 support.

This feature will support the reading of the v3 presentation api. Manifests saved locally using Cantaloupe will still be saved in v2.

PR #1619, opened 2025-06-13
Support iiif v.3, #1614

njkim:
Update the manifest manager and iiif viewer to support presentation v3.

Note: this work should eventually be moved either to core or to the [IIIF Arches App](https://github.com/archesproject/arches/issues/12279) assuming depending on when we are able to accomplish creating such an app.


arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
...
+            if (Array.isArray(val) && returnFirstVal) val = val[0];
+        } else if (version === 3) {
+            if (i18n) val = val[i18n];
+            if (Array.isArray(val) && returnFirstVal) val = val[0];
+        }
+        return val;
+    };
+    const getCanvases = (manifestData) => {
+        const canvases = [];
+        if (getVersion(manifestData) === 3) {
```
chiatt: It's probably just a bit better to check the version once at the top of the function rather than in each conditional statement.
```suggestion
        const version = getVersion(manifestData)
        if (version === 3) {
```

arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
...
+                        canvas.text = canvas.label;
+                        canvases.push(canvas);
+                    });
+                }
+            });
+        };
+        return canvases;
+    };
+    const getCanvas = (manifestData, canvasId, updateCanvas) => {
+        if (getVersion(manifestData) === 3) {
```
chiatt: Same here - get the version just once at the top of the function.

arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
+define([], function() {
+    const getVersion = (manifestData) => {
+        const urlString = manifestData?.['@context'];
+        const url = new URL(urlString);
+        if (url.pathname.split("/")[3].startsWith("3")) {
+            return 3;
+        } else if (url.pathname.split("/")[3].startsWith("2")) {
+            return 2;
+        }
```
chiatt, line 11:
```suggestion
        } else {
            throw new Error("Unable to identify version of IIIF presentation api. ");
        }
```

arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
...
+                let canvasIndex = 0;
+                if (sequence.canvases.length > 0) {
+                    if (!updateCanvas) {
+                        canvasIndex = sequence.canvases.findIndex((c) => (getCanvasService(c) === canvasId));
+                    }
+                    return sequence.canvases[canvasIndex];
+                }    
+            }
+        }
+        return null;
```
chiatt: Without a return statement, the function would return `undefined` instead of `null`. Would that suffice? If so maybe we could remove this return statement.

arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
...
+            return label;
+        }
+    };
+    const getCanvasService = (canvas) => {
+        if (canvas.images) {
+            return canvas.images[0].resource.service['@id'];
+        } else if (canvas.items) {
+            return canvas.items[0].items[0].body[0].service[0]["@id"];
+        };
+        return null;
```
chiatt: Maybe remove and let return undefined?

arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
...
+                const value = {};
+                Object.entries(data).forEach(([k, v]) => {
+                    value[k] = v['en'][0];
+                });
+                return value;
+            });
+        };
+    };
+    // I need a better name for this function
+    const changeCanvas = (manifestData) => {
```
chiatt: Maybe?
```suggestion
    const getInitialCanvas = (manifestData) => {
```

arches_for_science/media/js/utils/iiif-utils.js
```diff
@@ -0,0 +1,158 @@
...
+        getManifestDataValue,
+        getManifestLabel,
+        getManifestId,
+        getManifestThumbnail,
+        getCanvases,
+        getCanvas,
+        getCanvasLabel,
+        getCanvasService,
+        getMetadata,
+        changeCanvas,
```
chiatt:
```suggestion
        getInitialCanvas,
```

arches_for_science/media/js/views/components/plugins/manifest-manager.js
```diff
@@ -236,7 +237,7 @@ define([
 
             this.manifestData.subscribe(function(manifestData) {
                 if (manifestData) {
-                    self.selectCanvas(manifestData.sequences[0].canvases[0]);
+                    self.selectCanvas(iiifUtils.changeCanvas(manifestData));
```
chiatt:
```suggestion
                    self.selectCanvas(iiifUtils.getInitialCanvas(manifestData));
```

chiatt (changes requested):
Just a couple minor changes. Overall this is a really nice improvement. Moving a lot of that iiif logic into a utility file makes the workflows a bit easier to read.

chiatt (approved):

PR #1623, opened 2025-06-19
Update physical thing materials branch and reference data

chiatt:
Updates the aat thesaurus with chemical elements and ceramic wares and the Physical thing materials branch to support recording of chemical elements


njkim (approved):
The chemical elements are loaded and linked to the physical things and the pottery concept, too. All look good.

PR #1630, opened 2025-06-26
Add edtf node to phys thing timespan prod evt

chiatt:
Add edtf node to phys thing timespan prod evt and update concepts and collections


njkim (approved):
lgtm, the package loaded without any issue.

PR #1631, opened 2025-06-26
Add symbol to element label

chiatt:
Adds the chemical symbol to the element prefLabel and adds insert of 'pottery - visual works' into the Object Types - Physical Thing collection


njkim (approved):
lgtm

PR #1633, opened 2025-06-26
Fix thumbnail loading, #1632

### LINKED ISSUE #1632: Showing the search result thumbnails for digital resource often takes long.
The thumbnail in the search result sometimes takes long time especially when it fails.

njkim:
Fix the thumbnail error by getting the correct url from the arrays of ids.


chiatt (approved):
lgtm!

PR #1638, opened 2025-06-30
Prevent error when creating a digital resource for a canvas

### LINKED ISSUE #1637: Creation of digital resource fails on canvas image upload
An error prevents the creation of a new manifest when a digital resource is created for a manifest

chiatt:


njkim (approved):
lgtm

PR #1640, opened 2025-07-01
Add command to update manifest domain

chiatt:
Adds a command to allow an admin to perform a find and replace on a manifest domain after migrating between domains.


aarongundel (approved):
👏 lGtm!

PR #1641, opened 2025-07-01
Prevent duplication of a manifest on creation

chiatt:
Prevent duplicate manifest from getting created when the manifest url includes the domain.


njkim (approved):
lgtm

PR #1643, opened 2025-07-02
Fix iiif v3 label and thumbnail in workflows, #1642

### LINKED ISSUE #1642: V3 presentation api not working in analysis area workflow
It's working fine for Sample Taking, but errors on the image step of Analysis Area

njkim:


chiatt (approved):
Looks good!

PR #1651, opened 2025-07-07
minimum term search length should be one for elements

### LINKED ISSUE #1649: Allow users to search by single letter chemical symbols
Users need to be able to search for resources based on chemical symbol. Some symbols, however, have only one letter and Arches does not offer any concept search suggestions based on one letter. So, for concepts (not terms), users need to be able to get concept search suggestions based on a single letter.

aarongundel:
closes #1649


chiatt (approved):
LGTM

PR #1653, opened 2025-07-08
Ensure valueids in reference files match ids in migration

### LINKED ISSUE #1652: Valueids in package do not match those in concept migration
The valueids of recent concepts and collections do not match those generated by migrations that load concepts and collections for existing system. This could result in a divergence between extant and new instances of AFS. The ids in the package should updated to match the migration.

chiatt:
fixes #1652


aarongundel (approved):
LGTM

PR #1655, opened 2025-07-16
Add observation and instrument data for project reporting

chiatt:
Add observation and instrument data for project reporting


aarongundel (approved):
LGTM

PR #1663, opened 2025-08-02
Rename 3d workflows

chiatt:
Renames 3d workflows in a migration per https://github.com/archesproject/afs-demo/issues/41


aarongundel (approved):
@chiatt I thought this was just specific to the demo - but all good.

PR #1674, opened 2026-01-30
Upgrade AfS to 8.1 - replace arches urls

njkim:


njkim:
Close this in favor of #1675

